PR #8, opened 2021-05-24
Create index entries

cendhu:
This commit creates index entries whenever a block commit happens

### DIFF
--- a/internal/worldstate/leveldb/commit_and_query.go
+++ b/internal/worldstate/leveldb/commit_and_query.go
@@ -158,6 +158,11 @@ func (l *LevelDB) GetConfig() (*types.ClusterConfig, *types.Metadata, error) {
 	return config, metadata, nil
 }
 
+// GetIndexDefinition returns the index definition of a given database
+func (l *LevelDB) GetIndexDefinition(dbName string) ([]byte, *types.Metadata, error) {
+	return l.Get(worldstate.DatabasesDBName, dbName)
+}
+
 // Commit commits the updates to the database
 func (l *LevelDB) Commit(dbsUpdates map[string]*worldstate.DBUpdates, blockNumber uint64) error {
 	for dbName, updates := range dbsUpdates {
--- a/internal/stateindex/index_entries.go
+++ b/internal/stateindex/index_entries.go
@@ -1,0 +1,299 @@
+// Copyright IBM Corp. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package stateindex
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"reflect"
+
+	"github.com/IBM-Blockchain/bcdb-server/internal/worldstate"
+	"github.com/IBM-Blockchain/bcdb-server/pkg/types"
+)
+
+// IndexDBPrefix is the prefix added to each user database to create an index
+// database for that user database
+const IndexDBPrefix = "_index_"
+
+func constructIndexEntries(updates map[string]*worldstate.DBUpdates, db worldstate.DB) (map[string]*worldstate.DBUpdates, error) {
+	indexEntries := make(map[string]*worldstate.DBUpdates)
+
+	for dbName, update := range updates {
+		indexDef, _, err := db.GetIndexDefinition(dbName)
+		if err != nil {
+			return nil, err
+		}
+
+		if indexDef == nil {
+			continue
+		}
+
+		index := map[string]types.Type{}
+		if err := json.Unmarshal(indexDef, &index); err != nil {
+			return nil, err
+		}
+
+		newIndexToBeCreated, oldIndexToBeDeleted, err := indexEntriesForWrites(update.Writes, index, db, dbName)
+		if err != nil {
+			return nil, err
+		}
+
+		toBeDeletedIndexEntries, err := indexEntriesForDeletes(update.Deletes, index, db, dbName)
+		if err != nil {
+			return nil, err
+		}
+		oldIndexToBeDeleted = append(oldIndexToBeDeleted, toBeDeletedIndexEntries...)
+
+		dbUpdates := &worldstate.DBUpdates{}
+		for _, ind := range newIndexToBeCreated {
+			dbUpdates.Writes = append(dbUpdates.Writes, &worldstate.KVWithMetadata{
+				Key: ind,
+			})
+		}
+		dbUpdates.Deletes = append(dbUpdates.Deletes, oldIndexToBeDeleted...)
+
+		if len(dbUpdates.Writes) > 0 || len(dbUpdates.Deletes) > 0 {
+			indexEntries[IndexDBPrefix+dbName] = dbUpdates
+		}
+	}
+
+	return indexEntries, nil
+}
+
+func indexEntriesForWrites(
+	writes []*worldstate.KVWithMetadata,
+	index map[string]types.Type,
+	db worldstate.DB,
+	dbName string,
+) ([]string, []string, error) {
+	newIndexEntries, err := indexEntriesForNewValues(writes, index)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var keysUpdated []string
+	for _, w := range writes {
+		keysUpdated = append(keysUpdated, w.Key)
+	}
+	existingIndexEntries, err := indexEntriesOfExistingValue(keysUpdated, index, db, dbName)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	newEntries, err := toStrings(newIndexEntries)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	existingEntries, err := toStrings(existingIndexEntries)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	newIndexToBeCreated, oldIndexToBeDeleted := removeDuplicateIndexEntries(newEntries, existingEntries)
+	return newIndexToBeCreated, oldIndexToBeDeleted, nil
+}
+
+func indexEntriesForDeletes(deletes []string, index map[string]types.Type, db worldstate.DB, dbName string) ([]string, error) {
+	existingIndexOfDeletedValues, err := indexEntriesOfExistingValue(deletes, index, db, dbName)
+	if err != nil {
+		return nil, err
+	}
+
+	return toStrings(existingIndexOfDeletedValues)
+}
+
+type indexEntry struct {
+	Attribute string      `json:"a"`
+	Type      types.Type  `json:"t"`
+	Value     interface{} `json:"v"`
+	Key       string      `json:"k"`
+}
+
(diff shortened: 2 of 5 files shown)


internal/blockprocessor/committer.go
```diff
@@ -346,21 +347,33 @@ func constructDBEntriesForDBAdminTx(tx *types.DBAdministrationTx, version *types
 			}
 			value = v
 
+			// for each DB, if index is defined, we need to create an
+			// index DB to store index entries for that DB
+			indexDB := &worldstate.KVWithMetadata{
```
tock-ibm, line 352: What would be the value associated with the index db "\_index\_dbName" ?
Are all indices included in the same "\_index\_dbName" under different key schemas?

internal/blockprocessor/committer.go
```diff
@@ -369,21 +382,44 @@ func constructDBEntriesForDBAdminTx(tx *types.DBAdministrationTx, version *types
...
+		}
+
+		// Only when the index definition is provided for the existing DB for the
+		// first time, we will reach here
+		indexDB := &worldstate.KVWithMetadata{
+			Key: "_index_" + dbName,
+			Metadata: &types.Metadata{
+				Version: version,
+			},
+		}
```
tock-ibm, line 414: What is the value?

internal/blockprocessor/committer.go
```diff
@@ -369,21 +382,44 @@ func constructDBEntriesForDBAdminTx(tx *types.DBAdministrationTx, version *types
 			value = v
 		}
 
-		db := &worldstate.KVWithMetadata{
+		createDB := &worldstate.KVWithMetadata{
 			Key:   dbName,
```
tock-ibm, line 386: I am confused... this loop iterates on index definitions for existing DBs, right? because you deleted index definitions for new DBs in the previous loop. Correct?
So this essentially is an update to the value of an existing DB right? 
So the variable name is confusing - `updateDBIndex` would be more clear, IMO.

internal/blockprocessor/committer.go
```diff
@@ -369,21 +382,44 @@ func constructDBEntriesForDBAdminTx(tx *types.DBAdministrationTx, version *types
...
 			Value: value,
 			Metadata: &types.Metadata{
 				Version: version,
 			},
 		}
-		indexForExistingDBs = append(indexForExistingDBs, db)
+		indexForExistingDBs = append(indexForExistingDBs, createDB)
```
tock-ibm, line 392: Questions:
1. If the database already contains an index, that would overwrite its definition and render the old one stale; maybe it makes more sense to delete it and start fresh? or if the update just adds an index, it makes sense to add keys to it;
2. For a database that already contains data, a new index starts as stale.

How do you plan to handle the index building process, in terms of determining when an index db is in synch with the database it refers to?

One idea is to have the "value" of the \_index\_ database (currently empty I see) to reflect its state - stale, in-synch, etc. Just an idea, you might have thought of a better one... but I would like to learn what the plan is...

internal/blockprocessor/committer.go
```diff
@@ -332,12 +332,13 @@ func constructDBEntriesForDataTx(tx *types.DataTx, version *types.Version) []*wo
 	return dbsUpdates
 }
 
-func constructDBEntriesForDBAdminTx(tx *types.DBAdministrationTx, version *types.Version) ([]*worldstate.DBUpdates, error) {
+func constructDBEntriesForDBAdminTx(tx *types.DBAdministrationTx, version *types.Version, db worldstate.DB) ([]*worldstate.DBUpdates, error) {
```
tock-ibm, line 335: No unit tests?

internal/blockprocessor/index_entries_test.go
```diff
@@ -0,0 +1,812 @@
...
+	lc := &logger.Config{
+		Level:         "debug",
+		OutputPath:    []string{"stdout"},
+		ErrOutputPath: []string{"stderr"},
+		Encoding:      "console",
+	}
+	logger, err := logger.New(lc)
+	require.NoError(t, err)
+
+	dir, err := ioutil.TempDir("/tmp", "index")
```
tock-ibm, line 37: The empty string "" instead of "/tmp" makes this platform independent

internal/blockprocessor/index_entries_test.go
```diff
@@ -0,0 +1,812 @@
...
+			},
+		},
+		// {
+		// 	name:  "deletes of existing kv pairs",
+		// 	setup: func(db worldstate.DB) {},
+		// },
+		// {
+		// 	name:  "write and delete of new and existing values",
+		// 	setup: func(db worldstate.DB) {},
+		// },
```
tock-ibm, line 361: Are those forgotten or TODOs?

internal/blockprocessor/index_entries.go
```diff
@@ -0,0 +1,284 @@
+// Copyright IBM Corp. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package blockprocessor
+
```
tock-ibm, line 4: I think all the index code - the stuff that implements the particular indexing algorithm we plan to use - should be in another package. 
I would also like to see some documentation on the interface, possibly taken from your presentation, to accompany the methods and tests below. These two files should probably be in separate PR, dedicated to index entries, separate from the index db creation, above. Right now it is really difficult to review without context.

internal/worldstate/leveldb/commit_and_query.go
```diff
@@ -158,6 +158,11 @@ func (l *LevelDB) GetConfig() (*types.ClusterConfig, *types.Metadata, error) {
 	return config, metadata, nil
 }
 
+// GetIndexDefinition returns the index definition of a given database
+func (l *LevelDB) GetIndexDefinition(dbName string) ([]byte, *types.Metadata, error) {
+	return l.Get(worldstate.DatabasesDBName, dbName)
+}
+
```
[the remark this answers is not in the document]
cendhu: Done.

internal/blockprocessor/committer.go
```diff
@@ -369,21 +382,44 @@ func constructDBEntriesForDBAdminTx(tx *types.DBAdministrationTx, version *types
...
+		// If the index definition is empty, we need to delete all index entries
+		// associated with that. This is simply done by deleteing the index DB
+		if dbIndex == nil || dbIndex.GetAttributeAndType() == nil {
+			if db.Exist("_index_" + dbName) {
+				toDeleteDBs = append(toDeleteDBs, "_index_"+dbName)
+			}
+			continue
+		}
+
+		if db.Exist("_index_" + dbName) {
```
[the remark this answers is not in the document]
cendhu: Done

internal/blockprocessor/committer.go
```diff
@@ -346,21 +347,33 @@ func constructDBEntriesForDBAdminTx(tx *types.DBAdministrationTx, version *types
 			}
 			value = v
 
+			// for each DB, if index is defined, we need to create an
+			// index DB to store index entries for that DB
+			indexDB := &worldstate.KVWithMetadata{
+				Key: "_index_" + dbName,
```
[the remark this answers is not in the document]
cendhu: Done

internal/blockprocessor/committer.go
cendhu: For now, there is a single index per database. For each database, there is an index database. To allow multiple indexes per database, later, we will introduce either namespaces or hash(index) as postfix for the indexDB name. Having a database per index to store index entries is very flexible to perform fine-grained index management (lock, dropping the whole database when the index is deleted, dropping the whole database when the index changes completely, etc...)

internal/blockprocessor/committer.go
cendhu: it is nil for now.

internal/blockprocessor/committer.go
cendhu: Yes, good catch. Done.

internal/blockprocessor/committer.go
cendhu: For this PR, we are handling only index definition creation/updation events. Further, as stated in this issue https://github.com/IBM-Blockchain/bcdb-server/issues/14, we are considering the database to be empty. To address your Q1, we have https://github.com/IBM-Blockchain/bcdb-server/issues/15 and for Q2, we have https://github.com/IBM-Blockchain/bcdb-server/issues/21

When the database is not empty and a new index is created or an existing index is updated, we need to build index entries in the background. There are multiple scenarios and kind of complex. For example, consider the following scenarios:

1. **Database is not empty and new index is being defined** -- we need to perform an open-ended range query on the whole database and build index entry in batch, write the index entries to the index database and have a pointer to record till what key we have processed the index. Further, there can be new block commits and new values might get added. We need to keep track of them (even during node failure) and consider them once the existing data are processed. Note that the range query in leveldb acquires a snapshot and new writes are not visible. I am ignoring many corner cases here to keep it brief. In short, we need to store (1) a flag to denote whether the DB is in sync, (2) if being synched, what is the last key processed for the existing data; (3) if being synced, new keys committed while synching the index entries (just a list of block number is adequate); (4) if synched, what is the last processed block.
2. **Database is not empty and the existing index is being updated while the existing index entries are in sync with the database**
3. **Database is not empty and the existing index is being updated while the existing index entries are not in sync with the database**

Scenarios 2 and 3 are complex. Hence, we need to maintain more states to see (1) whether the index entries are in sync; (2) whether node recovery after failure can restart where it left off, etc...

If you prefer to merge this PR after having this feature end-to-end, I am okay with it.

internal/blockprocessor/committer.go
cendhu: Added

internal/blockprocessor/index_entries_test.go
cendhu: Done

internal/blockprocessor/index_entries_test.go
cendhu: Removed commented code.

internal/blockprocessor/index_entries.go
cendhu: Agree. I have made this PR to introduce only index creation logic. I have pushed a separate PR for the index DB creation. I am not sure about the interfaces. The index package will provide a function constructIndexEntries() for updates. We are using only the composite key-based index.

internal/blockprocessor/committer.go
tock-ibm: >If you prefer to merge this PR after having this feature end-to-end, I am okay with it.

No, my goal is to understand the underlying assumptions of the first phase of implementation. Clearly the scenarios described above are complex and require a lot of work. hence I assume that you are shooting for a simpler milestone to begin with. I am ok with that, just want to learn what it is though.

internal/blockprocessor/committer.go
tock-ibm: OK, an index database per index definition sounds good. I understand that our first milestone is for a single index definition per db.

gennadylaventman (approved):

PR #11, opened 2021-05-25
Integrate transport into replicator and transaction processor

tock-ibm:


internal/bcdb/transaction_processor.go
```diff
@@ -43,78 +47,119 @@ type transactionProcessor struct {
...
 	go p.txReorderer.Start()
 	p.txReorderer.WaitTillStart()
 
 	go p.blockCreator.Start()
 	p.blockCreator.WaitTillStart()
 
-	p.blockReplicator.Start()
+	p.peerTransport.Start() // Starts internal goroutine
+
+	p.blockReplicator.Start() // Starts internal goroutine
```
cendhu, line 162: Should we do a wait till start?

internal/blockprocessor/processor.go
```diff
@@ -56,6 +56,20 @@ func New(conf *Config) *BlockProcessor {
 	}
 }
 
+// Bootstrap initializes the ledger and database with the first block, which contains a config transaction.
+// This block is a.k.a. the "genesis block".
+func (b *BlockProcessor) Bootstrap(configBlock *types.Block) error {
+	if err := b.recoverWorldStateDBIfNeeded(); err != nil {
+		panic(errors.WithMessage(err, "error while recovering node"))
+	}
```
cendhu, line 64: I am not sure whether this is needed. We would reach here only if the block height is 0. When it is 0, all other stores must be empty.

cendhu (commented):
LGTM. Some unit tests seem to be missing. Further, the PR requires a rebase.

internal/bcdb/transaction_processor.go
tock-ibm: This method waits until ready internally, we should probably refactor the others to do the same as well....

internal/blockprocessor/processor.go
tock-ibm: You have a point here... I added it because there some failures during tests, probably for the init of the trie... so I added both. Looking into the recover shows it makes no senses because indeed the ledger height and store height will bothe be 0.  

Nevertheless, looking deeper, I think I did not pay enough attention to the difference between:
```
	blockStoreHeight, err := b.blockStore.Height()
	stateDBHeight, err := b.committer.db.Height()
```
and when calling the `Bootstrap()` method in the `transaction_processor.go` I get the store height and not the ledger height:
```

	ledgerHeight, err := conf.db.Height()
	if err != nil {
		return nil, err
	}
	if ledgerHeight == 0 {
		p.logger.Info("Bootstrapping the ledger and database")
		tx, err := PrepareBootstrapConfigTx(conf.config)
		if err != nil {
			return nil, err
		}
		bootBlock, err := blockcreator.BootstrapBlock(tx)
		if err != nil {
			return nil, err
		}
		if err = p.blockProcessor.Bootstrap(bootBlock); err != nil {
			return nil, err
		}
	}
```
I changed that to `ledgerHeight, err := conf.blockStore.Height()`.

internal/blockprocessor/processor.go
```diff
@@ -56,6 +56,16 @@ func New(conf *Config) *BlockProcessor {
 	}
 }
 
+// Bootstrap initializes the ledger and database with the first block, which contains a config transaction.
+// This block is a.k.a. the "genesis block".
+func (b *BlockProcessor) Bootstrap(configBlock *types.Block) error {
+	if err := b.initAndRecoverStateTrieIfNeeded(); err != nil {
+		panic(errors.WithMessage(err, "error while recovering node state trie"))
```
cendhu, line 63: `ValidateAndCommit()` panics because it is executing within a goroutine. For `Bootstrap()`, we can return the error.

internal/bcdb/transaction_processor.go
```diff
@@ -43,78 +47,119 @@ type transactionProcessor struct {
...
+			BlockOneQueueBarrier: p.blockOneQueueBarrier,
+			BlockStore:           conf.blockStore,
+			ProvenanceStore:      conf.provenanceStore,
+			StateTrieStore:       conf.stateTrieStore,
+			DB:                   conf.db,
+			Logger:               conf.logger,
+		},
+	)
+
+	ledgerHeight, err := conf.blockStore.Height()
```
cendhu, line 92: Good catch.

cendhu (approved):

cendhu:
Merging is blocked. @tock-ibm can you look at signing commits?

internal/blockprocessor/processor.go
tock-ibm: done

cendhu (approved):

PR #27, opened 2021-05-27
Add With() to logger

tock-ibm:

### DIFF
--- a/pkg/logger/logger_test.go
+++ b/pkg/logger/logger_test.go
@@ -5,6 +5,7 @@ package logger
 import (
 	"io/ioutil"
 	"os"
+	"path"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


pkg/logger/logger_test.go
```diff
@@ -298,3 +298,114 @@ func TestErrorPath(t *testing.T) {
...
+
+	tests := []struct {
+		level               string
+		fileName            string
+		expectedMessages    []string
+		notExpectedMessages []string
+	}{
+		{
+			level:    "debug",
+			fileName: "/tmp/debug.txt",
```
gennadylaventman, line 320: Can we use here some temp location and not go direct to /tmp folder

pkg/logger/logger_test.go
```diff
@@ -298,3 +298,114 @@ func TestErrorPath(t *testing.T) {
...
+			expectedMessages: []string{
+				"debug message is logged\t{\"id\": \"node1\"}",
```
cendhu, line 322: is it possible to prefix the `{"id": "node1"}` with the log message?

tock-ibm: not really... or no that I know of, it's a feature of zap

pkg/logger/logger_test.go
tock-ibm: Done

PR #54, opened 2021-06-09
Order preserving encoding of +ve/-ve value of indexed attribute

cendhu:
When the indexed attribute value is a number, it can be
either a positive number or a negative number. Further, we
need to store these numbers in the indexDB while preserving
the order (i.e., 1 < 5) which is needed for performing range query
on the indexed attributes.

Hence, this commit identifies whether the value of an index attribute
is +ve or -ve and accordingly encoding them to preserve the order.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.15
 require (
 	github.com/cayleygraph/cayley v0.7.7
 	github.com/cayleygraph/quad v1.1.0
+	github.com/gogo/protobuf v1.3.0
 	github.com/golang/protobuf v1.4.2
 	github.com/golang/snappy v0.0.1
 	github.com/google/uuid v1.1.1
--- a/internal/stateindex/encoding.go
+++ b/internal/stateindex/encoding.go
@@ -1,0 +1,83 @@
+package stateindex
+
+import (
+	"math"
+)
+
+// encodeOrderPreservingVarUint64 returns a byte-representation for a uint64 number such that
+// all zero-bits starting bytes are trimmed in order to reduce the length of the array
+// For preserving the order in a default bytes-comparison, first byte contains the number of remaining bytes.
+func encodeOrderPreservingVarUint64(n uint64) []byte {
+	var bytePosition int
+	for bytePosition = 0; bytePosition <= 7; bytePosition++ {
+		if byte(n>>(56-(bytePosition*8))) != 0x00 {
+			break
+		}
+	}
+
+	size := int8(8 - bytePosition)
+	encodedBytes := make([]byte, size+1)
+	encodedBytes[0] = byte(size)
+	p := 1
+	for r := bytePosition; r <= 7; r++ {
+		encodedBytes[p] = byte(n >> (56 - (r * 8)))
+		p++
+	}
+	return encodedBytes
+}
+
+// decodeOrderPreservingVarUint64 decodes the number from the bytes obtained from method 'encodeOrderPreservingVarUint64'.
+// It returns the decoded number.
+func decodeOrderPreservingVarUint64(b []byte) uint64 {
+	size := int(b[0])
+	var v uint64
+	for i := 7; i >= 8-size; i-- {
+		v = v | uint64(b[size-(7-i)])<<(56-(i*8))
+	}
+	return v
+}
+
+// encodeReverseOrderVarUint64 returns a byte-representation for a uint64 number such that
+// the number is first subtracted from MaxUint64 and then all the leading 0xff bytes
+// are trimmed and replaced by the number of such trimmed bytes. This helps in reducing the size.
+// In the byte order comparison this encoding ensures that encodeReverseOrderVarUint64(A) > encodeReverseOrderVarUint64(B),
+// If B > A
+func encodeReverseOrderVarUint64(n uint64) []byte {
+	n = math.MaxUint64 - n
+
+	var bytePosition int
+	for bytePosition = 0; bytePosition <= 7; bytePosition++ {
+		if byte(n>>(56-(bytePosition*8))) != 0xff {
+			break
+		}
+	}
+
+	size := int8(8 - bytePosition)
+	encodedBytes := make([]byte, size+1)
+	encodedBytes[0] = byte(bytePosition)
+	p := 1
+	for r := bytePosition; r <= 7; r++ {
+		encodedBytes[p] = byte(n >> (56 - (r * 8)))
+		p++
+	}
+	return encodedBytes
+}
+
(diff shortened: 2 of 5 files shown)


internal/stateindex/encoding.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	sizeBytes := proto.EncodeVarint(uint64(size))
+	if len(sizeBytes) > 1 {
+		panic(fmt.Errorf("[]sizeBytes should not be more than one byte because the max number it needs to hold is 8. size=%d", size))
+	}
+	encodedBytes := make([]byte, size+1)
+	encodedBytes[0] = sizeBytes[0]
+	copy(encodedBytes[1:], bytes[startingIndex:])
+	return encodedBytes
+}
```
tock-ibm, line 37: I think there is a faster way to do this.... instead of triple allocation and double copy:
-  implement the loop over the uint64 bytes as they do inside the `binary.BigEndian.PutUint64(bytes, number)` function, (shifting the value by multiples of 8 every round) 
- when you find the first non-zero byte allocate the right size of slice (size+1), 
- the varint encoding of the uint64 size is just byte(size)
- populate the first two bytes, and continue the loop writing one byte at a time

internal/stateindex/encoding.go
```diff
@@ -0,0 +1,96 @@
...
+		return 0, 0, errors.Errorf("decoded size (%d) from DecodeVarint is more than available bytes (%d)", s, len(bytes)-1)
+	default:
+		// no error
+		size := int(s)
+		decodedBytes := make([]byte, 8)
+		copy(decodedBytes[8-size:], bytes[1:size+1])
+		numBytesConsumed := size + 1
+		return binary.BigEndian.Uint64(decodedBytes), numBytesConsumed, nil
+	}
+}
```
tock-ibm, line 59: Again, since the size varint is always one byte long, you can get the code snippet from `proto.DecodeVarint` that deals with the right range... e.g.

```
v = uint64(b[0])
if v < 0x80 {
	return v, 1
}
return <error>
```
and you don't need to allocate `decodedBytes` because you can easily implement the code inside "binary.BigEndian.Uint64(decodedBytes)` by looping around input argument `bytes`.

Nit: bytes is a name of a package, maybe call the input argument  `b`?

internal/stateindex/encoding.go
```diff
@@ -0,0 +1,96 @@
...
+	numFFBytes := int(s)
+	decodedBytes := make([]byte, 8)
+	realBytesNum := 8 - numFFBytes
+	copy(decodedBytes[numFFBytes:], bytes[1:realBytesNum+1])
+	numBytesConsumed := realBytesNum + 1
+	for i := 0; i < numFFBytes; i++ {
+		decodedBytes[i] = 0xff
+	}
+	return (math.MaxUint64 - binary.BigEndian.Uint64(decodedBytes)), numBytesConsumed
+}
```
tock-ibm, line 96: same efficiency comments as above for the non-reverse order

internal/stateindex/encoding_test.go
```diff
@@ -0,0 +1,39 @@
...
+			t.Fatalf("A smaller integer should result into smaller bytes. Encoded bytes for [%d] is [%x] and for [%d] is [%x]",
+				i, i+1, value, nextValue)
+		}
+		decodedValue, _, err := decodeOrderPreservingVarUint64(value)
+		require.NoError(t, err, "Error via calling DecodeOrderPreservingVarUint64")
+		if decodedValue != uint64(i) {
+			t.Fatalf("Value not same after decoding. Original value = [%d], decode value = [%d]", i, decodedValue)
+		}
+	}
+}
```
tock-ibm, line 24: check also numbers that have no leading zeros...

internal/stateindex/encoding_test.go
```diff
@@ -0,0 +1,39 @@
...
+		if !(bytes.Compare(value, nextValue) > 0) {
+			t.Fatalf("A smaller integer should result into greater bytes. Encoded bytes for [%d] is [%x] and for [%d] is [%x]",
+				i, i+1, value, nextValue)
+		}
+		decodedValue, _ := decodeReverseOrderVarUint64(value)
+		if decodedValue != uint64(i) {
+			t.Fatalf("Value not same after decoding. Original value = [%d], decode value = [%d]", i, decodedValue)
+		}
+	}
+}
```
tock-ibm, line 39: check numbers that have no leading 0xFF bytes

tock-ibm (commented):
Partial review, will continue with `index_entries.go` later....

internal/stateindex/index_entries_test.go
```diff
@@ -422,6 +429,9 @@ func TestIndexEntriesOfExistingValues(t *testing.T) {
 		"age": types.Type_NUMBER,
 	}
 
+	encoded25 := encodeOrderPreservingVarUint64(uint64(25))
+	encoded26 := encodeOrderPreservingVarUint64(uint64(26))
```
tock-ibm, line 433: I would also check the corner cases of 0, +/- max-int, and verify that the ordering of a mix of positive and negative numbers is as expected.

tock-ibm (commented):
Completed my review. 
Overall looks good but some efficiency and testing comments.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.15
 require (
 	github.com/cayleygraph/cayley v0.7.7
 	github.com/cayleygraph/quad v1.1.0
+	github.com/gogo/protobuf v1.3.0
```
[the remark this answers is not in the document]
cendhu: `Goimports` automatically picked this one.

internal/stateindex/encoding.go
cendhu: Done.

internal/stateindex/encoding.go
cendhu: Done

internal/stateindex/encoding_test.go
cendhu: Done

internal/stateindex/encoding_test.go
cendhu: Done

internal/stateindex/index_entries_test.go
cendhu: Partially Done. I have created issue #145 I can add the order testing only after https://github.com/IBM-Blockchain/bcdb-server/pull/131 gets merged. I would in fact add this test in the query executor side as well.

tock-ibm (approved):

tock-ibm:
Looks good to me.

PR #69, opened 2021-06-14
Fix files according to goimports

C0rWin:
Running goimports on the project files to fix formatting issues.


cendhu (approved):
Thanks, @C0rWin

PR #82, opened 2021-06-20
Consensus metadata proto

tock-ibm:
Add another field to the block to hold consensus metadata.
This fields holds consensus specific data that must be persisted with the block.
For Raft we need to save the term and index in order to restart the node and handle snapshots.


protos/block_and_transaction.proto
```diff
@@ -217,3 +219,12 @@ enum Type {
   STRING = 1;
   BOOLEAN = 2;
 }
+
+// ConsensusMetadata holds data specific to the consensus protocol ordering the block.
+// The field prefix indicated the protocil used, e.g. "raft_*".
```
cendhu, line 224: We can use the `one-of` feature when we add more protocol.

cendhu (approved):

PR #83, opened 2021-06-20
Raft config extension

tock-ibm:
Add maxInflightBlocks and snapshotIntervalSize to raft config.


protos/configuration.proto
```diff
@@ -99,6 +99,12 @@ message RaftConfig {
   // pass between heartbeats. That is, a leader sends heartbeat
   // messages to maintain its leadership every HeartbeatTick ticks.
   uint32 heartbeat_ticks = 3;
+
+  // Limits the max number of in-flight blocks (i.e. Raft messages).
+  uint32 max_inflight_blocks = 4;
+
+  // Take a snapshot when cumulative data since last snapshot exceeds a certain size in bytes.
+  uint64 snapshot_interval_size = 5;
```
cendhu, line 107: can't we use the `types.RaftConfig` instead of `RaftConfig` in the `shared_config.go`? I remember asking this question in the past and don't remember your response.

config/shared_config.go
```diff
@@ -59,6 +59,10 @@ type RaftConf struct {
 	// pass between heartbeats. That is, a leader sends heartbeat
 	// messages to maintain its leadership every HeartbeatTick ticks.
 	HeartbeatTicks uint32
+	// Limits the max number of in-flight blocks (i.e. Raft messages).
+	MaxInflightBlocks uint32
```
cendhu, line 63: Does the proposer wait to cut a block once this limit is reached?

tock-ibm: This is an internal raft parameter, if the number of in flight messages is beyond this, the leader will stop proposing. I think it will lead to the proposing go-routine to block. 

If the propose go-routine is blocked, it will not pull blocks from the channel that connects the replication and the block creator, which will eventually cause the block creator to stop. But all that in the next PR, not now...

protos/configuration.proto
tock-ibm: Actually, if I recall correctly, you asked me specifically not to use the `type.*` protobuf messages in the config package?
It used to be like that (reused protos in config) but I changed it when I did the shared config PR a while pack...
There is a duplication of  structs between `config.*` and `types.*` in many other levels as well...

cendhu: I see. Maybe the reason was to expand the `config.yml` independently of the proto message.

cendhu (approved):

PR #84, opened 2021-06-20
Integrate raft node

tock-ibm:
Import the etcd/raft lib module
Write the main loop that operates the RSM;
- Order(block)
- Deliver(block)

Assumptions:
- No dynamic config changes to membership
- Use (as module) the WAL from etcd 
- Use (copy) `Storage` from Fabric 
- No snapshots: don't create snapshots, don't accept snapshot
- Support node creation
- Support node shutdown and restart
- Test single node only

### DIFF
--- a/internal/replication/blockreplicator.go
+++ b/internal/replication/blockreplicator.go
@@ -5,8 +5,10 @@ package replication
 
 import (
 	"context"
-	"errors"
+	"github.com/golang/protobuf/proto"
 	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/IBM-Blockchain/bcdb-server/config"
 	"github.com/IBM-Blockchain/bcdb-server/internal/comm"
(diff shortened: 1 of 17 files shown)


internal/replication/blockreplicator.go
```diff
@@ -5,163 +5,451 @@ package replication
...
+		case <-raftTicker.C:
+			if status := br.raftNode.Status().String(); status != raftStatusStr {
+				br.lg.Debugf("Raft node status: %+v", status)
+				raftStatusStr = status
+			}
+			br.raftNode.Tick()
+
+		case rd := <-br.raftNode.Ready():
+			startStoring := time.Now()
+			if err := br.raftStorage.Store(rd.Entries, rd.HardState, rd.Snapshot); err != nil {
```
cendhu, line 215: is this a sync write?

internal/replication/storage.go
```diff
@@ -0,0 +1,432 @@
...
+	"github.com/pkg/errors"
+	"go.etcd.io/etcd/etcdserver/api/snap"
+	"go.etcd.io/etcd/pkg/fileutil"
+	"go.etcd.io/etcd/raft"
+	"go.etcd.io/etcd/raft/raftpb"
+	"go.etcd.io/etcd/wal"
+	"go.etcd.io/etcd/wal/walpb"
+)
+
+// MaxSnapshotFiles defines max number of etcd/raft snapshot files to retain
```
cendhu, line 27: storage layer is taken from `fabric/orderer/consensus/etcdraft/storage.go` Hence, I read the code but not doing a deep review.

cendhu (approved):

PR #90, opened 2021-06-24
change proto messages to make it fit for cURL

cendhu:
Currently, we do double marshaling of response
to a query. This does not print a human readable
format on the screen. Hence, this commit avoids
usage of bytes field in the response. Further,
marshaling is done only to put signature on the
response. Overall, it will help to improve the
performance slightly and make cURL usable.


protos/response.proto
```diff
@@ -9,76 +9,164 @@ package types;
...
 
 message GetTxIDsSubmittedByResponse {
-  repeated string txIDs = 1;
+  ResponseHeader header = 1;
+  repeated string txIDs = 2;
 }
 
-message TxResponse {
-  TxReceipt receipt = 1;
+message TxReceiptResponseEnvelope {
```
tock-ibm, line 164: This one is missing a comment like all the rest have 
(OCD)

protos/response.proto
```diff
@@ -9,76 +9,164 @@ package types;
...
 
 //========= Part II Provenance API responses
 
+// GetBlock
+message GetBlockResponseEnvelope {
+  GetBlockResponse response = 1;
+  bytes signature = 2;
+}
+
 message GetBlockResponse {
```
tock-ibm, line 82: maybe rename both to `GetBlockHeader*` ?

tock-ibm (approved):

tock-ibm:
Looks good to me. All renaming suggestions could be ignored or addresses in a follow-up issue, as this one is to big to review again.

cendhu:
Let's merge this @tock-ibm I will submit another PR to address your comments.

PR #91, opened 2021-06-24
use lower case characters in keys for the json request/response

cendhu:
For JSON documents, we need to use lower case characters in keys to maintain consistency. Hence, this commit makes changes to the fields in proto messages to ensure consistent naming for the keys.


protos/block_and_transaction.proto
```diff
@@ -72,13 +72,13 @@ message UserAdministrationTxEnvelope {
...
 message DataTx {
-  repeated string mustSignUserIDs = 1;
-  string txID = 2;
-  repeated DBOperation DBOperations = 3;
+  repeated string must_sign_user_ids = 1;
+  string tx_id = 2;
+  repeated DBOperation db_operations = 3;
 }
 
 message DBOperation {
```
gennadylaventman, line 80: Just one question - shouldn't we change `DBOperation`, `DBAdministrationTx`, `DBIndex`, `GetDBStatusQuery` change to same way as we changed fields names?

gennadylaventman (commented):
Shouldn't we change `types.*` type names to same format? For example `DBOperation` to `DbOperation`.

protos/block_and_transaction.proto
cendhu: No, we don't need to change the message name. Only the field names are used when we do `json.Marshal()`

gennadylaventman (approved):

PR #102, opened 2021-07-03
Fix missing return of the error in mptrie

C0rWin:
One of the previous commits missed returning the error within Update API if the empty key was put as a parameter to the function call. 

Another issue is within `GetProof` the error was compared with itself, i.e.

```
	if err != err {
		return nil, err
	}
```

which obviously always false. This commit fixes it as well. 


gennadylaventman (approved):
Looks good to me.

PR #103, opened 2021-07-03
Use variable type while asserting within swith case

C0rWin:
Within the implementation of mptie there were a few nits, while using a switch to assert node type there is a way to leverage type knowledge to eliminate the need for casting variables later on. 


gennadylaventman (approved):
Looks good to me

PR #104, opened 2021-07-04
Add access to MPTrie data

gennadylaventman:
Added support for data proof to `ledger_request_handler.go` and `http.go` from REST side and to `db.go` and `ledger_query_procesor.go` from internal implementation side.


internal/httphandler/ledger_request_handler.go
```diff
@@ -36,14 +36,18 @@ func NewLedgerRequestHandler(db bcdb.DB, logger *logger.SugarLogger) http.Handle
...
+	// HTTP GET "/ledger/proof/tx/{blockId}?idx={idx}" gets proof for tx with idx index inside block blockId
 	handler.router.HandleFunc(constants.GetTxProof, handler.txProof).Methods(http.MethodGet).Queries("idx", "{idx:[0-9]+}")
+	// HTTP GET "/ledger/proof/data/{blockId}/{dbname}/{key}?deleted={true|false}" gets proof for tx with idx index inside block blockId
+	handler.router.HandleFunc(constants.GetDataProof, handler.dataProof).Methods(http.MethodGet).Queries("deleted", "{deleted:true|false}")
 	// HTTP GET "/ledger/tx/receipt/{txId}" gets transaction receipt
 	handler.router.HandleFunc(constants.GetTxReceipt, handler.txReceipt).Methods(http.MethodGet)
 	// HTTP GET "/ledger/path?start={startId}&end={endId}" with invalid query params
 	handler.router.HandleFunc(constants.GetPath, handler.invalidPathQuery).Methods(http.MethodGet)
-	// HTTP GET "/ledger/proof/{blockId}?idx={idx}" with invalid query params
+	// HTTP GET "/ledger/proof/tx/{blockId}?idx={idx}" with invalid query params
```
tock-ibm, line 47: This will break SDK, need to reflect it there as well

internal/httphandler/ledger_request_handler.go
```diff
@@ -36,14 +36,18 @@ func NewLedgerRequestHandler(db bcdb.DB, logger *logger.SugarLogger) http.Handle
 	handler.router.HandleFunc(constants.GetBlockHeader, handler.blockQuery).Methods(http.MethodGet)
 	// HTTP GET "/ledger/path?start={startId}&end={endId}" gets shortest path between blocks
 	handler.router.HandleFunc(constants.GetPath, handler.pathQuery).Methods(http.MethodGet).Queries("start", "{startId:[0-9]+}", "end", "{endId:[0-9]+}")
-	// HTTP GET "/ledger/proof/{blockId}?idx={idx}" gets proof for tx with idx index inside block blockId
+	// HTTP GET "/ledger/proof/tx/{blockId}?idx={idx}" gets proof for tx with idx index inside block blockId
 	handler.router.HandleFunc(constants.GetTxProof, handler.txProof).Methods(http.MethodGet).Queries("idx", "{idx:[0-9]+}")
+	// HTTP GET "/ledger/proof/data/{blockId}/{dbname}/{key}?deleted={true|false}" gets proof for tx with idx index inside block blockId
```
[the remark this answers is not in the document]
tock-ibm: Again, why blockID and not key version?

internal/httphandler/ledger_request_handler.go
```diff
@@ -36,14 +36,18 @@ func NewLedgerRequestHandler(db bcdb.DB, logger *logger.SugarLogger) http.Handle
...
+	handler.router.HandleFunc(constants.GetDataProof, handler.dataProof).Methods(http.MethodGet).Queries("deleted", "{deleted:true|false}")
```
tock-ibm, line 42: Is it possible to use this with a default `deleted=false` ?
deleted seems to be rare enough so that we allow a default for the frequent case.

pkg/constants/http.go
```diff
@@ -36,7 +36,8 @@ const (
 	LedgerEndpoint = "/ledger/"
 	GetBlockHeader = "/ledger/block/{blockId:[0-9]+}"
 	GetPath        = "/ledger/path"
-	GetTxProof     = "/ledger/proof/{blockId:[0-9]+}"
+	GetTxProof     = "/ledger/proof/tx/{blockId:[0-9]+}"
+	GetDataProof   = "/ledger/proof/data/{blockId:[0-9]+}/{dbname:" + `[0-9a-zA-Z_\-\.]+` + "}/{key}"
```
tock-ibm, line 40: I think that if the user make a request to `/ledger/proof/tx` `/ledger/proof/tx/` `/ledger/proof/data` or `/ledger/proof/data/` he will get an unformatted error response content from within `gorilla/mux` which is not interpreted very well by the SDK.

internal/bcdb/db.go
```diff
@@ -60,6 +60,9 @@ type DB interface {
 	// GetTxProof returns intermediate hashes to recalculate merkle tree root from tx hash
 	GetTxProof(userID string, blockNum uint64, txIdx uint64) (*types.GetTxProofResponseEnvelope, error)
 
+	// GetDataProof returns hashes path from value to root in merkle-patricia trie
+	GetDataProof(userID string, blockNum uint64, dbname string, key string, deleted bool) (*types.GetDataProofResponseEnvelope, error)
```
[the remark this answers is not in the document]
gennadylaventman: Interesting idea. Let me explain original idea - block number used as time marker. So, once user ask for proof for key K inside block N, it doesn't mean that this key was updated in block N, it just mean that at the time of block N creation, value associated with key K was value with hash located at the end of proof.

internal/httphandler/ledger_request_handler.go
gennadylaventman: Sure, next step will update sdk. Lets tag it after that to 0.1.1

internal/httphandler/ledger_request_handler.go
gennadylaventman: Have to thinks about it...

pkg/constants/http.go
gennadylaventman: It will get default 404 response. I can add multiple invalid handlers to take care of following cases: 
1. ` /ledger/proof/tx`
2. `/ledger/proof/data`
3. `/ledger/proof/data/{blockId}`
4. `ledger/proof/data/{blockId}/{dbname}`
But once user uses sdk, no such thing is possible. wdut?

tock-ibm: I think it makes sense to cover all the corners, after all, the REST spec is public and user/community can write her own SDK.... especially in other languages, like Java, Node, who knows...

internal/bcdb/db.go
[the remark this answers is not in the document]
cendhu: @gennadylaventman I think it should be the version as mentioned by @tock-ibm For example, multiple transactions in a block can blindly write to a KV pair. The end value of the KV pair is the value written by the last transactions. However, the provenance store would return all writes performed to the KV pair. In order to get a proof for any data, the user can pass the block number and the version. 

Now, I wonder whether we take care of all blind writes within a block. I think, we do not. Am I right?

internal/httphandler/ledger_request_handler.go
```diff
@@ -36,14 +36,18 @@ func NewLedgerRequestHandler(db bcdb.DB, logger *logger.SugarLogger) http.Handle
 	handler.router.HandleFunc(constants.GetBlockHeader, handler.blockQuery).Methods(http.MethodGet)
 	// HTTP GET "/ledger/path?start={startId}&end={endId}" gets shortest path between blocks
 	handler.router.HandleFunc(constants.GetPath, handler.pathQuery).Methods(http.MethodGet).Queries("start", "{startId:[0-9]+}", "end", "{endId:[0-9]+}")
-	// HTTP GET "/ledger/proof/{blockId}?idx={idx}" gets proof for tx with idx index inside block blockId
+	// HTTP GET "/ledger/proof/tx/{blockId}?idx={idx}" gets proof for tx with idx index inside block blockId
```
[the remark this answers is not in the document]
gennadylaventman: Done

internal/httphandler/ledger_request_handler.go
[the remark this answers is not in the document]
gennadylaventman: Done

internal/httphandler/ledger_request_handler.go
gennadylaventman: Done

internal/httphandler/ledger_request_handler.go
gennadylaventman: Because sdk call to same `constants.URLTxProof(blockNum, txIndex)` function, new tag after merge will be enough. In any case, opened new issue to update sdk: https://github.com/IBM-Blockchain/bcdb-sdk/issues/29

internal/bcdb/db.go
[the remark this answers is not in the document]
gennadylaventman: We are not, because it will produce multiple trie roots for same block, at least as number of tx in block. It is doable, but will require same change it integration code and will increase block header size. We had discussion with @tock-ibm  and for now, at least, we going with block number. I already have issue to store metadata hash with value, adding this, I think, will solve version problem.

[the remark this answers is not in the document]
cendhu: How about we allow only one write to a key per block? In general, blind writes are very rare. Even if blind writes are done by the user, if we allow only one blind write to a key per block, we can prove all states present in the provenance store. Having a root hash per transaction would complicate things more.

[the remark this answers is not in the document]
gennadylaventman: I don't think root hash per tx is related to amount of writes to same key in block because each write to any key updates root hash. If we handle tx as one atomic update, each tx with writes updates root hash. This is why I mentioned root hash per tx. We had discussion with Yoav and we decided to keep current approach, just to store metadata hash with value hash and to make following changes to REST API `ledger/proof/data/db/key?block={blockNum}&deleted={true|false}`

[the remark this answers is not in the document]
cendhu: Discussed offline. We will make changes in MVCC to allow only one write to a given key per block.

internal/bcdb/ledger_query_procesor.go
```diff
@@ -118,6 +122,50 @@ func (p *ledgerQueryProcessor) getProof(userId string, blockNum uint64, txIdx ui
...
+func (p *ledgerQueryProcessor) getDataProof(userId string, blockNum uint64, dbname string, key string, deleted bool) (*types.GetDataProofResponse, error) {
+	hasAccess, err := p.identityQuerier.HasLedgerAccess(userId)
+	if err != nil {
+		return nil, err
+	}
+
+	if !hasAccess {
+		return nil, &interrors.PermissionErr{ErrMsg: fmt.Sprintf("user %s has no permission to access the ledger", userId)}
+	}
+	block, err := p.blockStore.Get(blockNum)
```
cendhu, line 134: `GetHeader()` instead of `Get()`.

internal/bcdb/ledger_query_procesor.go
```diff
@@ -118,6 +122,50 @@ func (p *ledgerQueryProcessor) getProof(userId string, blockNum uint64, txIdx ui
 	}, nil
 }
 
+func (p *ledgerQueryProcessor) getDataProof(userId string, blockNum uint64, dbname string, key string, deleted bool) (*types.GetDataProofResponse, error) {
```
cendhu, line 125: `isDeleted` instead of `deleted`

internal/bcdb/ledger_query_procesor_test.go
```diff
@@ -330,6 +354,84 @@ func constructProvenanceEntriesForDataTx(tx *types.DataTx, version *types.Versio
 	return txpData
 }
 
+func applyBlockOnStateTrie(trie *mptrie.MPTrie, worldStateUpdates []*worldstate.DBUpdates) error {
```
cendhu, line 357: As this is an exact duplicate of the same function in the `committer.go`, we can extract the common part of an exported function and then call it here and in the committer package

internal/bcdb/ledger_query_procesor_test.go
```diff
@@ -330,6 +354,84 @@ func constructProvenanceEntriesForDataTx(tx *types.DataTx, version *types.Versio
...
+		dbsUpdates[i] = &worldstate.DBUpdates{
+			DBName:  ops.DbName,
+			Writes:  kvWrites,
+			Deletes: kvDeletes,
+		}
+	}
+
+	return dbsUpdates
+}
+
```
cendhu, line 434: export the same function in `committer.go` and call it here.

internal/httphandler/ledger_request_handler.go
```diff
@@ -36,14 +36,27 @@ func NewLedgerRequestHandler(db bcdb.DB, logger *logger.SugarLogger) http.Handle
...
 	handler.router.HandleFunc(constants.GetTxReceipt, handler.txReceipt).Methods(http.MethodGet)
 	// HTTP GET "/ledger/path?start={startId}&end={endId}" with invalid query params
 	handler.router.HandleFunc(constants.GetPath, handler.invalidPathQuery).Methods(http.MethodGet)
-	// HTTP GET "/ledger/proof/{blockId}?idx={idx}" with invalid query params
+	// HTTP GET "/ledger/proof/tx/{blockId}?idx={idx}" with invalid query params
+	handler.router.HandleFunc(constants.GetTxProofPrefix, handler.invalidTxProof).Methods(http.MethodGet)
+	// HTTP GET "/ledger/proof/tx/{blockId}?idx={idx}" with invalid query params
 	handler.router.HandleFunc(constants.GetTxProof, handler.invalidTxProof).Methods(http.MethodGet)
+	// HTTP GET "/ledger/proof/data/{blockId}/{dbname}/{key}" with invalid query params
+	handler.router.HandleFunc(constants.GetDataProofPrefix, handler.invalidDataProof).Methods(http.MethodGet)
```
cendhu, line 55: Not related to this PR. We need to do this for other handlers too.

pkg/constants/http_test.go
```diff
@@ -78,7 +78,21 @@ func TestURLConstruction(t *testing.T) {
...
 			},
-			expectedURL: "/ledger/proof/1?idx=2",
+			expectedURL: "/ledger/proof/tx/1?idx=2",
+		},
+		{
+			name: "URLDataProof deleted false",
+			execute: func() string {
+				return URLDataProof(1, "db1", "key", false)
+			},
+			expectedURL: "/ledger/proof/data/db1/key?block=1",
```
cendhu, line 88: I am okay with the usage of `block` rather than `block_number`. To maintain consistency, we need to use `block` for other HTTP requests too. I think any of us can address this in a separate PR.

protos/query.proto
```diff
@@ -92,6 +92,19 @@ message GetTxProofQueryEnvelope {
   bytes signature = 2;
 }
 
+message GetDataProofQuery {
+  string user_id = 1;
+  uint64 block_number = 2;
+  string db_name = 3;
+  string key = 4;
+  bool deleted = 5;
```
cendhu, line 100: `is_deleted`

cendhu (commented):
A few minor comments. Otherwise, LGTM.

pkg/constants/http_test.go
gennadylaventman: Agreed. I will open issue and lets discuss it.

https://github.com/IBM-Blockchain/bcdb-server/issues/111

cendhu (approved):
Thanks, @gennadylaventman

PR #108, opened 2021-07-11
Updating crypto material and docker image generation

gennadylaventman:
Crypto materials and configuration folders structure changed to folder structure used by car demo application
Crypto generation now done by dockerized openssl - see cryptoGen.sh

To create minimal crypto configuration, you can run:
```
./scripts/cryptoGen.sh sampleconfig
```
To create crypto configuration for more users/server, you can run:
```
./scripts/cryptoGen.sh sampleconfig <extra users>
```
For example, in case you want to generate crypto materials for extra users used in car demo, you can run:
```
./scripts/cryptoGen.sh sampleconfig alice bob dmv dealer
```

To generate docker image, after you generated crypto materials, run:
```
make docker
```

To invoke bcdb docker container, you can run:
```
docker run -it --rm -v $(pwd)/sampleconfig/:/etc/bcdb-server -p 6001:6001 -p 7050:7050 bcdb-server
``` 

### DIFF
--- a/sampleconfig/config/config.yml
+++ b/sampleconfig/config/config.yml
@@ -8,9 +8,9 @@ server:
     # of the node. This certificate and associated key is used in all
     # client facing communication, as well as for signing blocks and
     # transaction replies.
-    certificatePath: /etc/bcdb-server/pki/node/node.cert
+    certificatePath: /etc/bcdb-server/crypto/node/node.pem
     # identity.keyPath denotes the path to the private key of the node.
-    keyPath: /etc/bcdb-server/pki/node/node.key
+    keyPath: /etc/bcdb-server/crypto/node/node.key
   # The listen address and port of the network interface used for client
   # communication. The external address (or host name) of this interface
   # must be accessible to clients, and is declared in:
@@ -19,7 +19,7 @@ server:
   # where consensus.members[i].nodeID == node.identity.id
   network:
     # network.address denotes the listen address
-    address: 127.0.0.1
+    address:
     # network.port denotes the listen port
     port: 6001
   database:
@@ -28,7 +28,7 @@ server:
     name: leveldb
     # database.ledgerDirectory denotes the root path
     # where we store all ledger data
-    ledgerDirectory: ledger
+    ledgerdirectory: /etc/bcdb-server/ledger
   queueLength:
     # queueLength.transaction denotes the maximum
     # queue length of waiting transactions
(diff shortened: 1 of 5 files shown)


sampleconfig/config/config.yml
```diff
@@ -28,7 +28,7 @@ server:
     name: leveldb
     # database.ledgerDirectory denotes the root path
     # where we store all ledger data
-    ledgerDirectory: ledger
+    ledgerdirectory: /etc/bcdb-server/database
```
cendhu, line 31: let's use `ledger` instead of `database`

scripts/carDemoGen.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
+
+if [ -z "$1" ]
+  then
+    echo "carDemoGen.sh folder"
+    exit 1
+fi
+printf "http://127.0.0.1:6001" > $1/server.url
+mkdir $1/txs
+
```
cendhu, line 10: why the car demo is part of the server?

scripts/cryptoGen.sh
```diff
@@ -1,29 +1,51 @@
...
+
+rm -rf "$BASE_DIR/crypto"
 
 echo "Creating PKIs folders"
-mkdir -p "$BASE_DIR/pki/ca"
+mkdir -p "$BASE_DIR/crypto/CA"
 
 echo "Generate root CA private key"
-openssl ecparam -name prime256v1 -genkey -noout -out "$BASE_DIR/pki/ca/rootCA.key"
+openssl ecparam -name prime256v1 -genkey -noout -out "$BASE_DIR/crypto/CA/CA.key"
```
cendhu, line 33: Does this script require significant change when we move to cluster? Like one CA, multiple nodes, multiple users?

scripts/cryptoGenDocker.sh
```diff
@@ -0,0 +1,55 @@
+#! /bin/bash
```
cendhu, line 1: The usage pattern is confusing for me. Why do we have two `cryptogen` scripts. For users, having one would avoid confusion.

sampleconfig/config/config.yml
gennadylaventman: ok

scripts/carDemoGen.sh
gennadylaventman: Actually, it is only end2end example we have now. I had discussion with Yoav and he suggested make docker image be able to support car demo out of the box. But I will move this script to sdk.

scripts/cryptoGen.sh
gennadylaventman: Multiple users already supported. Multiple servers can be added quite easy. In cluster environment it should be splitted to 3 scripts - one for CA generation, one for users and one for nodes.

scripts/cryptoGenDocker.sh
gennadylaventman: Ok. It started with one script - `cryptoGen.sh`. But, because some differences in Mac, it was failing and required extra configuration of openssl on Mac. So, I created script that use dockerized openssl, but kept original script as well. If you think we don't need both, I can remove original script.

scripts/cryptoGen.sh
gennadylaventman: If you will look in `create_pki` function - it creates private key and certificate for some entity - it can be user or node and sign it with CA certificate. We can use it as base for any script that creates crypto for any amount of users/nodes.

scripts/cryptoGenDocker.sh
cendhu: In that case, rather than using two scripts, we can use a single script and pass the machine architecture as an input. Otherwise, we can just use `docker` script alone removing the old script. I leave the choice to you.

cendhu (approved):

PR #113, opened 2021-07-15
fix dbupdates

cendhu:
If we use one dbupdates for each tx in a block, we would do one sync write per tx.
As it degrades the performance, this PR makes one dbUpdates per database. As a result,
the number of sync writes would be equal to the number of dbs' data
modified in the block.

With this PR, the time taken to commit a block to worldstate reduced
from 150 msecs to 6 msecs

### DIFF
--- a/internal/bcdb/ledger_query_procesor_test.go
+++ b/internal/bcdb/ledger_query_procesor_test.go
@@ -6,17 +6,16 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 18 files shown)


gennadylaventman (commented):
Maybe I missed it, but I didn't find test that actually checks if all DBUpdates from multiple transactions are located together.

cendhu:
> Maybe I missed it, but I didn't find test that actually checks if all DBUpdates from multiple transactions are located together.

Good catch. Updated the unit test.

gennadylaventman (approved):

PR #124, opened 2021-07-22
Improving MPTrie performance

gennadylaventman:
Improving MPTrie.persistsSubtrie performance.

### DIFF
--- a/internal/mptrie/mptrie.go
+++ b/internal/mptrie/mptrie.go
@@ -454,33 +454,39 @@ func (t *MPTrie) Commit(blockNum uint64) error {
 	t.lock.RLock()
 	defer t.lock.RUnlock()
 
-	if err := t.persistSubtrie(t.root); err != nil {
+	rootHash, err := t.root.hash()
+	if err != nil {
+		return err
+	}
+
+	if err := t.persistSubtrie(rootHash); err != nil {
 		return err
 	}
 	return t.store.CommitChanges(blockNum)
 }
 
-func (t *MPTrie) persistSubtrie(n TrieNode) error {
-	wasChanged, err := t.persistNode(n)
+func (t *MPTrie) persistSubtrie(nodePtr []byte) error {
+	wasChanged, err := t.persistNode(nodePtr)
 	if err != nil {
 		return err
 	}
+
 	// If node wasn't changed, no need to persist iт child nodes and can return
 	if !wasChanged {
 		return nil
 	}
-	switch n := n.(type) {
+	node, err := t.store.GetNode(nodePtr)
+	if err != nil {
+		return err
+	}
+	switch n := node.(type) {
 	case *BranchNode:
 		for _, childPtr := range n.Children {
-			if childPtr != nil {
-				child, err := t.store.GetNode(childPtr)
-				if err != nil {
-					return err
-				}
-
-				if err = t.persistSubtrie(child); err != nil {
-					return err
-				}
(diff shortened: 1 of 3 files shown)


internal/mptrie/store/open.go
```diff
@@ -14,6 +14,8 @@ import (
 )
 
 var (
+	nodeGetTotal = 0
+	nodeGetDB = 0
```
cendhu, line 18: Please remove. For time, we can calculate and print the time taken. For low level metrics like these, we need to use prometheus.

internal/mptrie/store/update_and_query.go
```diff
@@ -171,6 +171,9 @@ func (s *Store) CommitChanges(blockNum uint64) error {
 	if err := s.trieDataDB.Write(batch, &opt.WriteOptions{Sync: true}); err != nil {
 		return err
 	}
+	s.logger.Infof("mptrie commit - total get ops %d, db get ops %d, put ops %d, persist ops %d", nodeGetTotal, nodeGetDB, len(s.inMemoryNodes), len(s.nodesToPersist))
+	nodeGetTotal = 0
+	nodeGetDB = 0
```
cendhu, line 176: Please remove. You can add time taken to persistTrie but for low level metrics, let's not use logs.

internal/mptrie/store/update_and_query.go
```diff
@@ -118,7 +120,6 @@ func (s *Store) PersistNode(nodePtr []byte) (bool, error) {
 	nodeBytesWithType, ok := s.inMemoryNodes[key]
 	if ok {
 		s.nodesToPersist[key] = nodeBytesWithType
-		delete(s.inMemoryNodes, key)
```
cendhu, line 121: why should we remove this line?

cendhu (commented):
I have tested the performance using this PR. The time taken to commit the trie is reduced by 72%. Nice work. 

A few minor comments.

internal/mptrie/mptrie.go
```diff
@@ -454,31 +454,36 @@ func (t *MPTrie) Commit(blockNum uint64) error {
...
 	case *BranchNode:
 		for _, childPtr := range n.Children {
 			if childPtr != nil {
-				child, err := t.store.GetNode(childPtr)
-				if err != nil {
-					return err
-				}
-
-				if err = t.persistSubtrie(child); err != nil {
+				if err = t.persistSubtrie(childPtr); err != nil {
```
C0rWin, line 486: While you are already at it, can you please reverse the logic of if statement above to reduce code indentation?

internal/mptrie/store/open.go
gennadylaventman: ok

internal/mptrie/store/update_and_query.go
gennadylaventman: ok

internal/mptrie/store/update_and_query.go
gennadylaventman: It doesn't affect anything - we will clear this map in `CommitChanges` call.

internal/mptrie/mptrie.go
gennadylaventman: Done, but it adds extra code line :)

gennadylaventman: Done

cendhu (approved):

PR #126, opened 2021-07-23
Parallel batch preparation in provenance store

cendhu:


internal/provenance/commit_and_query.go
```diff
@@ -99,39 +97,39 @@ func (s *Store) Commit(blockNum uint64, txsData []*TxDataForProvenance) error {
...
 		s.logger.Debugf("txID[%s]---(reads)--->value[%s]", tx.TxID, quad.NativeOf(value))
-		cayleyTx.AddQuad(quad.Make(tx.TxID, READS, value, ""))
+		batch.WriteQuad(quad.Make(tx.TxID, READS, value, ""))
 	}
 
 	return nil
 }
 
-func (s *Store) addWrites(tx *TxDataForProvenance, cayleyTx *graph.Transaction, commitSet map[string]string) error {
+func (s *Store) addWrites(tx *TxDataForProvenance, batch graph.BatchWriter) error {
```
gennadylaventman, line 132: Short question - why we pass to `*graph.Transaction`, but `graph.Batch` - pointer in first case and object itself in second?

cendhu: `graph.BatchWriter` is an interface while `graph.Transaction` is a struct.

gennadylaventman: Thanks.

gennadylaventman (approved):

PR #132, opened 2021-07-28
Update index entries structure

cendhu:
To enable logical operations such as <, >, <=, >=, we introduce
two additional fields in the index entry: ValuePosition and
KeyPosition.

Further, we unexport certain fields/functions and export some
which are to be used by query executor package

### DIFF
--- a/internal/stateindex/index_entries.go
+++ b/internal/stateindex/index_entries.go
@@ -13,12 +13,20 @@ import (
 	"github.com/IBM-Blockchain/bcdb-server/pkg/types"
 )
 
-// IndexDBPrefix is the prefix added to each user database to create an index
-// database for that user database
 const (
-	IndexDBPrefix  = "_index_"
-	PositiveNumber = "p"
-	NegativeNumber = "n"
+	// indexDBPrefix is the prefix added to each user database to create an index
+	// database for that user database
+	indexDBPrefix = "_index_"
+	// positiveNumber and negativeNumber are metadata that denote whether the value being
+	// index is positive or negative
+	positiveNumber = "p"
+	negativeNumber = "n"
+)
+
+const (
+	Beginning = iota
+	Existing
+	Ending
 )
 
 func constructIndexEntries(updates map[string]*worldstate.DBUpdates, db worldstate.DB) (map[string]*worldstate.DBUpdates, error) {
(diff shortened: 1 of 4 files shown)


internal/stateindex/index_entries.go
```diff
@@ -109,16 +117,19 @@ func indexEntriesForDeletes(deletes []string, index map[string]types.Type, db wo
...
-	Type      types.Type  `json:"t"`
-	Metadata  string      `json:"m"`
-	Value     interface{} `json:"v"`
-	Key       string      `json:"k"`
+// IndexEntry hold metadata associated with the attribute being indexed along with the attribute value and key
+type IndexEntry struct {
+	Attribute     string      `json:"a"`
+	Type          types.Type  `json:"t"`
+	Metadata      string      `json:"m"`
+	ValuePosition int8        `json:"vp"` // ValuePosition is used such that range query for lesser than, greater than can be executed easily
```
tock-ibm, line 125: Why?

internal/stateindex/index_entries.go
```diff
@@ -109,16 +117,19 @@ func indexEntriesForDeletes(deletes []string, index map[string]types.Type, db wo
...
+	Value         interface{} `json:"v"`
+	KeyPosition   int8        `json:"kp"` // KeyPosition is used such that range query for lesser than, greater than can be executed easily
```
tock-ibm, line 127: Why?

internal/stateindex/index_entries.go
```diff
@@ -13,12 +13,20 @@ import (
...
+	// positiveNumber and negativeNumber are metadata that denote whether the value being
+	// index is positive or negative
+	positiveNumber = "p"
+	negativeNumber = "n"
+)
+
+const (
+	Beginning = iota
+	Existing
+	Ending
```
tock-ibm, line 29: Could you say a few words on how this is used and why is it needed? 
It is difficult to understand at this stage.... This will probably become clearer in future commits... nevertheless, it would also be easier for the reader if there would be a comment that describes the index entry structure...
Why the iota? I only see `Existing` used in the code so far. Is there an intention to add more values, or to use `Beginning`  and `Ending`?

tock-ibm (approved):

tock-ibm:
The code looks good, but I have a few questions and requests for better documentation of the `IndexEntry` structure. This could be done in the next commit though.

PR #134, opened 2021-07-28
Range query plan

cendhu:


internal/queryexecutor/range_query_planner.go
```diff
@@ -0,0 +1,117 @@
...
+		// TODO: support number values
+		return nil, errors.New("numbers are not supported currently")
+	}
+
+	for c := range conds.conditions {
+		if c == constants.NotEqual {
+			// TODO: support not equal
+			return nil, errors.New(constants.NotEqual + " is not supported currently")
+		}
+	}
```
tock-ibm, line 32: Please also comment on that in the previous PR that dissects the query... I was asking my self about this when I read it.

Although I think extending the plan could be easy - check that the NE value is between begin & end, add it ad a list, do the range query, and then filter the results using the NE list, right?

I am OK with doing this in a different PR though.

internal/queryexecutor/range_query_planner.go
```diff
@@ -0,0 +1,117 @@
...
+			Attribute: attribute,
+			Type:      conds.valueType,
+		},
+	}
+
+	if len(conds.conditions) == 1 {
+		for c, v := range conds.conditions {
+			setPlanForSingleCondition(c, v, p)
+			return p, nil
+		}
```
tock-ibm, line 49: If `len==1`, isn't it clearer to just access `conds.conditions[0]` ?

internal/queryexecutor/range_query_planner.go
```diff
@@ -0,0 +1,117 @@
...
+type rangeQueryPlan struct {
+	startKey *stateindex.IndexEntry
+	endKey   *stateindex.IndexEntry
+}
+
+func createQueryPlan(attribute string, conds *attrConditions) (*rangeQueryPlan, error) {
+	// we assume this function to get only valid conditions
+	//   - eq and no other conditions
+	//   - lt and lte do not appear together
+	//   - gt and gte do not appear together
```
tock-ibm, line 20: It also makes sense to put that comment on the commit before this one, that validates the conditions...

internal/queryexecutor/range_query_planner.go
```diff
@@ -0,0 +1,117 @@
...
+func setPlanForSingleCondition(c string, v interface{}, p *rangeQueryPlan) {
+	switch c {
+	case constants.Equal:
+		p.startKey.ValuePosition = stateindex.Existing
+		p.startKey.Value = v
+		p.startKey.KeyPosition = stateindex.Beginning
+
+		p.endKey.ValuePosition = stateindex.Existing
+		p.endKey.Value = v
+		p.endKey.KeyPosition = stateindex.Ending
```
tock-ibm, line 66: I can't figure out from this commit or the one before it what is the purpose of `ValuePosition` & `KeyPosition`...
Could you comment on those? it would then be easier to review...

internal/queryexecutor/range_query_planner.go
cendhu: If we ignore the `$neq` first and then apply only after fetching the results, it would be less performant. I am planning to construct multiple range query plans when there is a `$neq` and is more efficient. It will be done as part of https://github.com/IBM-Blockchain/bcdb-server/issues/169

internal/queryexecutor/range_query_planner.go
cendhu: condition is a map, that's why.

internal/queryexecutor/range_query_planner.go
cendhu: Done, already.

internal/queryexecutor/range_query_planner.go
cendhu: Added descriptions for each condition.

internal/queryexecutor/range_query_planner.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return p, nil
+}
+
+func setPlanForSingleCondition(c string, v interface{}, p *rangeQueryPlan) {
+	// for all index entries, we have set ValuePosition to the `Existing` and
+	// KeyPosition to the `Existing`. While creating range query plan, we use
+	// `Beginning`, `Existing`, and `Ending` where `Beginning` < `Existing` < `Ending`.
+	// We can use `Beginning` to denote `before a begining of a value` and `Ending`
+	// to denote `till after the end`.
```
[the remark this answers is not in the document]
cendhu: Done.

internal/queryexecutor/range_query_planner.go
```diff
@@ -0,0 +1,142 @@
...
+		p.endKey.KeyPosition = stateindex.Ending
+	case constants.QueryOpGreaterThan:
+		// By setting startKey.KeyPosition to `Ending`, we instruct the executor
+		// to start scanning from the next greater value of `v`
+		p.startKey.ValuePosition = stateindex.Existing
+		p.startKey.Value = v
+		p.startKey.KeyPosition = stateindex.Ending
+
+		// By setting endKey.ValuePosition to `Ending`, we instruct the executor
+		// to scan till the end of entries
```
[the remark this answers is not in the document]
cendhu: Done

internal/queryexecutor/range_query_planner.go
```diff
@@ -0,0 +1,142 @@
...
+		p.endKey.ValuePosition = stateindex.Ending
+	case constants.QueryOpGreaterThanOrEqual:
+		// By setting startKey.KeyPosition to `Beginning`, we instruct the executor
+		// to start scanning from the beginning of the key that matches value `v`
+		p.startKey.ValuePosition = stateindex.Existing
+		p.startKey.Value = v
+		p.startKey.KeyPosition = stateindex.Beginning
+
+		// By setting endKey.ValuePosition to `Ending`, we instruct the executor
+		// to scan till the end of entries
```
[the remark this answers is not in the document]
cendhu: Done

internal/queryexecutor/range_query_planner.go
```diff
@@ -0,0 +1,142 @@
...
+		p.startKey.ValuePosition = stateindex.Beginning
+
+		// By setting endKey.KeyPosition to `Beginning`, we instruct the executor
+		// to scan till before the begining of the key that matches value `v`
+		p.endKey.ValuePosition = stateindex.Existing
+		p.endKey.Value = v
+		p.endKey.KeyPosition = stateindex.Beginning
+	case constants.QueryOpLesserThanOrEqual:
+		// By setting startKey.ValuePosition to `Beginning`, we instruct the executor
+		// to scan from the beginning of entries
```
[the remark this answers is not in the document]
cendhu: Done

internal/queryexecutor/range_query_planner.go
```diff
@@ -0,0 +1,142 @@
...
+		p.startKey.ValuePosition = stateindex.Existing
+		p.startKey.Value = v
+		p.startKey.KeyPosition = stateindex.Beginning
+
+		// By setting endKey.ValuePosition to `Ending`, we instruct the executor
+		// to scan till the end of entries
+		p.endKey.ValuePosition = stateindex.Ending
+	case constants.QueryOpLesserThan:
+		// By setting startKey.ValuePosition to `Beginning`, we instruct the executor
+		// to scan from the beginning of entries
```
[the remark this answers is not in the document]
cendhu: Done

internal/queryexecutor/range_query_planner_test.go
```diff
@@ -0,0 +1,277 @@
...
+				},
+			},
+			expectedPlan: &rangeQueryPlan{
+				startKey: &stateindex.IndexEntry{
+					Attribute:     "attr1",
+					Type:          types.Type_STRING,
+					Metadata:      "",
+					ValuePosition: stateindex.Existing,
+					Value:         "value-a",
+					KeyPosition:   stateindex.Beginning,
```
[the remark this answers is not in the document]
cendhu: agree but let's do that in a separate PR as it would affect unit-test in state index package too.

tock-ibm (approved):

PR #137, opened 2021-07-29
Take snapshots & restart from snapshot

tock-ibm:
This commit:
    - Adds code that allows each single node to create a stream of snapshots based on accumulated size criteria.
    - The last 4 snapshots are saved, older snapshots are purged.
    - The WAL is also pruned (compacted) when a snapshot is created.
    - A single node can restart from its own snapshot.

    However:
    - Catch up between leader and follower not supported yet. Therefore,
    - For a cluster scenario use snapshot-inteval that will never cause a snapshot, e.g. max-int64.

### DIFF
--- a/internal/replication/blockreplicator.go
+++ b/internal/replication/blockreplicator.go
@@ -22,6 +22,13 @@ import (
 	"go.etcd.io/etcd/wal"
 )
 
+const (
+	// DefaultSnapshotCatchUpEntries is the default number of entries
+	// to preserve in memory when a snapshot is taken. This is for
+	// slow followers to catch up.
+	DefaultSnapshotCatchUpEntries = uint64(4)
+)
+
 type BlockLedgerReader interface {
 	Height() (uint64, error)
 	Get(blockNumber uint64) (*types.Block, error)
@@ -47,6 +54,12 @@ type BlockReplicator struct {
 	clusterConfig   *types.ClusterConfig
 	lastKnownLeader uint64
 	appliedIndex    uint64
+
+	// needed by snapshotting
+	sizeLimit        uint64 // SnapshotIntervalSize in bytes
+	accDataSize      uint64 // accumulative data size since last snapshot
+	lastSnapBlockNum uint64
+	confState        raftpb.ConfState // Etcdraft requires ConfState to be persisted within snapshot
 
 	lg *logger.SugarLogger
 }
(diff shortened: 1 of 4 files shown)


internal/replication/blockreplicator.go
```diff
@@ -285,6 +311,9 @@ func (br *BlockReplicator) deliverEntries(committedEntries []raftpb.Entry) bool
 				break
 			}
 
+			position = i
```
gennadylaventman, line 314: I wondering why you need both `position` and `i`. I assume it is because `break` in line 321. And this lead to next question - why `break` and not `continue`?

internal/replication/blockreplicator_test.go
```diff
@@ -466,3 +467,169 @@ func TestBlockReplicator_ReConfig(t *testing.T) {
...
+		require.NoError(t, err)
+
+		snapList2 := replication.ListSnapshots(env.conf.Logger, env.conf.LocalConf.Replication.SnapDir)
+		require.Equal(t, 4, len(snapList2))
+		t.Logf("Snap list after 2nd close: %v", snapList2)
+		require.Equal(t, snapList[3], snapList2[2])
+
+		//recreate - snapshot and two following entries
+		err = env.Restart()
+		require.NoError(t, err)
```
gennadylaventman, line 625: I think I missing something, but don't we need to check is `env` restarted from correct snapshot?

gennadylaventman (commented):
2 questions/comments.

internal/replication/blockreplicator.go
tock-ibm: By the time the for loop ends, `position` will carry the last element in `committedEntries` that delivers a block. Blocks are delivered in `EntryNormal` only at this point. There might be `EntryConfChange` elements in the array past the last block, and `position` avoids them.

The `break` breaks from the switch, which is effectively continues to the next loop iteration.

internal/replication/blockreplicator_test.go
tock-ibm: Good question. In general, the node recovers correctly if it can continue to function: i.e. elect a leader and agree on block order. If block order after recovery is messed up (e.g. trying to `Append` the same block twice or `Append` with a gap), the mock ledger will panic and the test will fail.

In order to ensure the above, I added a few block submissions after this part (second recovery), and check that they get to the ledger ok. 

I also added a few comments on each of the test cases to make it easier to read.

gennadylaventman (approved):

PR #143, opened 2021-08-03
Execute json query

cendhu:


internal/queryexecutor/executor.go
```diff
@@ -15,26 +16,90 @@ import (
...
-func NewWorldStateQueryExecutor(db worldstate.DB, l *logger.SugarLogger) *WorldStateQueryExecutor {
-	return &WorldStateQueryExecutor{
+func NewWorldStateJSONQueryExecutor(db worldstate.DB, l *logger.SugarLogger) *WorldStateJSONQueryExecutor {
+	return &WorldStateJSONQueryExecutor{
 		db:     db,
 		logger: l,
 	}
 }
 
+func (e *WorldStateJSONQueryExecutor) ExecuteQuery(dbName string, selector []byte) (map[string]bool, error) {
```
tock-ibm, line 31: I would be helpful to document here the structure of the json object expected to be in `selector`. That is, unless you plan to document it in a later PR on a higher layer. It is a map, insid a map, inside a map, not a trivial structure.... it is ok to do this in another PR.

tock-ibm (approved):

PR #149, opened 2021-08-09
Allow config updates to Admins & CAs

tock-ibm:
Dynamic config of ClusterConfig is currently not supported and causes panic.
This commit allows config changes to ClusterConfig.Admins and ClusterConfig.CertAuthConfig,
as those have no effect on the replication component.

The server will panic if ClusterConfig.Nodes or ClusterConfig.ConsensusConfig are changed.
This will change once dynamic reconfig of consensus related objects is implemented.

### DIFF
--- a/internal/replication/blockreplicator.go
+++ b/internal/replication/blockreplicator.go
@@ -5,7 +5,6 @@ package replication
 
 import (
 	"context"
-	"github.com/golang/protobuf/proto"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -16,6 +15,7 @@ import (
 	"github.com/IBM-Blockchain/bcdb-server/internal/queue"
 	"github.com/IBM-Blockchain/bcdb-server/pkg/logger"
 	"github.com/IBM-Blockchain/bcdb-server/pkg/types"
+	"github.com/golang/protobuf/proto"
 	"github.com/pkg/errors"
 	"go.etcd.io/etcd/raft"
 	"go.etcd.io/etcd/raft/raftpb"
@@ -309,8 +309,11 @@ func (br *BlockReplicator) deliverEntries(committedEntries []raftpb.Entry) bool 
 			}
 
 			if reConfig != nil {
-				// TODO support reconfig
-				br.lg.Panic("Unexpected update to ClusterConfig during normal entry")
+				clusterConfig := reConfig.(*types.ClusterConfig)
+				if err := br.updateClusterConfig(clusterConfig); err != nil {
+					// TODO support dynamic re-config
+					br.lg.Panicf("Failed to update to ClusterConfig during raft normal entry: error: %s", err)
+				}
 			}
 
 		case raftpb.EntryConfChange:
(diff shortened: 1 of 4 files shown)


internal/replication/blockreplicator.go
```diff
@@ -309,8 +309,11 @@ func (br *BlockReplicator) deliverEntries(committedEntries []raftpb.Entry) bool
 			}
 
 			if reConfig != nil {
-				// TODO support reconfig
-				br.lg.Panic("Unexpected update to ClusterConfig during normal entry")
+				clusterConfig := reConfig.(*types.ClusterConfig)
+				if err := br.updateClusterConfig(clusterConfig); err != nil {
+					// TODO support dynamic re-config
+					br.lg.Panicf("Unexpected update to ClusterConfig during normal entry: error: %s", err)
```
gennadylaventman, line 315: Can you rephrase log message?

internal/replication/env_test.go
```diff
@@ -36,6 +37,13 @@ var clusterConfig1node = &types.ClusterConfig{
 		Port:        nodePortBase + 1,
 		Certificate: []byte("bogus-cert"),
 	}},
+	Admins: []*types.Admin{&types.Admin{
+		Id:                   "admin",
+		Certificate:          []byte("something"),
+		XXX_NoUnkeyedLiteral: struct{}{},
+		XXX_unrecognized:     nil,
+		XXX_sizecache:        0,
```
gennadylaventman, line 45: Should we fill XXX_ fileds?

internal/replication/utils.go
```diff
@@ -25,3 +27,41 @@ func raftEntryString(e raftpb.Entry) string {
...
 		e.Term, e.Index, e.Type, len(e.Data), h.Sum64())
 }
+
+func ClassifyClusterReConfig(currentConfig, updatedConfig *types.ClusterConfig) (nodes bool, consensus bool, ca bool, admins bool) {
+	curNodes := currentConfig.GetNodes()
+	updNodes := updatedConfig.GetNodes()
+	if len(curNodes) != len(updNodes) {
+		nodes = true
+	} else {
+		for i, node := range curNodes {
```
gennadylaventman, line 37: Do we keep nodes order in list? Can't it be same nodes, but in different order?

cendhu: Yes, good point.

internal/replication/blockreplicator.go
tock-ibm: I hope this is better"
`br.lg.Panicf("Failed to update to ClusterConfig during raft normal entry: error: %s", err)`

Anyway, this log message is temporary until dynamic consensus config is implemented...

internal/replication/env_test.go
tock-ibm: removed

internal/replication/utils.go
tock-ibm: Done, for both nodes and admins.

gennadylaventman (approved):

PR #151, opened 2021-08-10
Mptrie proof api moving to pkg

gennadylaventman:
Moving internal `mptrie.Proof` struct to pkg `state.Proof` with all relevant methods. 

To make it available to sdk.

### DIFF
--- a/pkg/state/proof.go
+++ b/pkg/state/proof.go
@@ -1,0 +1,116 @@
+package state
+
+import (
+	"bytes"
+
+	"github.com/IBM-Blockchain/bcdb-server/pkg/crypto"
+	"github.com/IBM-Blockchain/bcdb-server/pkg/types"
+	"github.com/pkg/errors"
+)
+
+var KeyDeleteMarkerBytes = []byte{1}
+
+type Proof struct {
+	// For each node in trie path, it contains bytes of all node fields and []byte{1} in case of deleted flag true
+	// path is from rom node contains value to root node
+	// Exactly same byte slices used to calculate node hash.
+	path []*types.MPTrieProofElement
+}
+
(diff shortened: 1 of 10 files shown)


pkg/state/proof.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"bytes"
+
+	"github.com/IBM-Blockchain/bcdb-server/pkg/crypto"
+	"github.com/IBM-Blockchain/bcdb-server/pkg/types"
+	"github.com/pkg/errors"
+)
+
+var DeleteBytes = []byte{1}
```
tock-ibm, line 11: This means "Key Deletion Marker" or "Key Deletion Tombstone" right? 
I suggest to rename to a more descriptive name...

pkg/state/proof.go
```diff
@@ -0,0 +1,116 @@
...
+			return false, err
+		}
+	}
+	// Check if calculated root hash if equal to supplied (stored in block)
+	return bytes.Equal(rootHash, hashToFind), nil
+}
+
+func (p *Proof) GetPath() []*types.MPTrieProofElement {
+	return p.Path
+}
```
tock-ibm, line 70: Why is this needed if Path is exported? 
maybe Path -> path?

tock-ibm (commented):
Looks good, had just a couple of nits...

pkg/state/proof.go
gennadylaventman: Ok

pkg/state/proof.go
gennadylaventman: Let me check in sdk

tock-ibm (approved):

PR #155, opened 2021-08-17
Catchup service

tock-ibm:
In this commit the comm component exposes a new endpoint on top
of the same http server that serves the Raft transport.

- /bcdb-peer is the base
- /bcdb-peer/height allows the other peer to request the ledger height
- /bcdb-peer/blocks allows the other peer to request a range of blocks

The blocks request uses query parameters to specify the range:

 /bcdb-peer/blocks?start=2&end=5

This will be used for state transfer from a leader to a follower that
is behind of the leader's most recent checkpoint. This PR only exposes
the service (server-side), the code that actually uses it when receiving
a snapshot trigger will be introduced in a following commit.


internal/comm/catchup_handler.go
```diff
@@ -0,0 +1,137 @@
...
+	w.Header().Set("Content-Type", mw.FormDataContentType())
+	for i, blockBytes := range blocks {
+		fw, err := mw.CreateFormFile(
+			fmt.Sprintf("block-%d", i),
+			fmt.Sprintf("num-%d", startBlockNum+uint64(i)))
+		if err != nil {
+			httputils.SendHTTPResponse(w, http.StatusInternalServerError, &types.HttpResponseErr{ErrMsg: err.Error()})
+			return
+		}
+		if _, err := fw.Write(blockBytes); err != nil {
```
gennadylaventman, line 113: Don't we need signatures here, to check sender originality?

internal/comm/catchup_handler.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (h *catchupHandler) heightRequest(w http.ResponseWriter, r *http.Request) {
+	h.lg.Debugf("height request: %s", r.URL)
+	height, err := h.ledgerReader.Height()
+	if err != nil {
+		httputils.SendHTTPResponse(w, http.StatusInternalServerError, &types.HttpResponseErr{ErrMsg: err.Error()})
+		return
+	}
+
+	httputils.SendHTTPResponse(w, http.StatusOK, HeightResponse{Height: height})
```
gennadylaventman, line 136: Signature?

internal/comm/catchup_handler_test.go
```diff
@@ -0,0 +1,211 @@
+// Copyright IBM Corp. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package comm_test
```
gennadylaventman, line 4: Maybe I miss something, but why do we need extra package for `comm_test`?

internal/comm/catchup_handler_test.go
```diff
@@ -0,0 +1,211 @@
...
+			require.NoError(t, err)
+			require.NotNil(t, blockBytes)
+
+			block := &types.Block{}
+			err = proto.Unmarshal(blockBytes, block)
+			require.NoError(t, err)
+			assert.Equal(t, bNum, block.Header.BaseHeader.Number)
+			bNum++
+		}
+	})
```
gennadylaventman, line 210: I miss one more test case: start < end, start < height, end > height

internal/comm/httptransport.go
```diff
@@ -5,6 +5,7 @@ package comm
 
 import (
 	"fmt"
+	"github.com/gorilla/mux"
```
gennadylaventman, line 8: goimports

internal/comm/httptransport.go
```diff
@@ -56,12 +60,14 @@ func NewHTTPTransport(config *Config) *HTTPTransport {
...
 
 	tr := &HTTPTransport{
-		logger:    config.Logger,
-		localConf: config.LocalConf,
-		stopCh:    make(chan struct{}),
-		doneCh:    make(chan struct{}),
+		logger:         config.Logger,
+		localConf:      config.LocalConf,
+		catchupHandler: NewCatchupHandler(config.Logger, config.LedgerReader),
+		router:         mux.NewRouter(),
```
gennadylaventman, line 66: Where do we use `router`?

internal/comm/catchup_handler.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if startBlockNum > height {
+		httputils.SendHTTPResponse(response, http.StatusBadRequest, &types.HttpResponseErr{fmt.Sprintf("requested startId [%d] is out of range, height is [%d]", startBlockNum, height)})
+		return
+	}
+
+	var blocks [][]byte
+	for i := startBlockNum; i <= endBlockNum; i++ {
+		if i > height {
+			break
```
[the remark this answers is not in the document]
tock-ibm: Yes

internal/comm/catchup_handler.go
tock-ibm: No, it is server to server, and I will handle authentication and authorization in the comm layer.
Basically mutual-TLS with the server verifying that the client belongs to the `Members` set, as defined in the cluster config.

internal/comm/catchup_handler.go
tock-ibm: See above

internal/comm/catchup_handler_test.go
tock-ibm: It is just that I prefer block-box testing, not white-box. I never access non exported vars or methods in these tests.

internal/comm/httptransport.go
tock-ibm: done

internal/comm/httptransport.go
tock-ibm: Good catch... not needed

internal/comm/catchup_handler_test.go
tock-ibm: done

internal/comm/catchup_handler.go
```diff
@@ -0,0 +1,137 @@
...
+			return
+
+		}
+		blockBytes, err := proto.Marshal(block)
+		if err != nil {
+			httputils.SendHTTPResponse(response, http.StatusInternalServerError, &types.HttpResponseErr{ErrMsg: err.Error()})
+			return
+
+		}
+		blocks = append(blocks, blockBytes)
```
cendhu, line 96: Not related to this PR but we need to add a `GetMarshaledBlock()` in the block store to avoid unmarshaling at block store and marshalling here.

cendhu (approved):

internal/comm/catchup_handler.go
gennadylaventman: Agreed. Opened issue https://github.com/IBM-Blockchain/bcdb-server/issues/170

PR #162, opened 2021-09-02
Refactor http utilities to common package

tock-ibm:
Moving 3 functions from `httphandler/utils.go` to a leaf package in `httputils/utils.go`.
These functions will be used in the next commit, that introduces the http catch-up service in the comm package.

No functional change.

### DIFF
--- a/internal/httphandler/utils.go
+++ b/internal/httphandler/utils.go
@@ -1,71 +1,27 @@
 // Copyright IBM Corp. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+
 package httphandler
 
 import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
-	"fmt"
-	"log"
 	"net/http"
 	"strconv"
 	"time"
 
-	internalerror "github.com/IBM-Blockchain/bcdb-server/internal/errors"
-
-	"github.com/IBM-Blockchain/bcdb-server/internal/bcdb"
+	"github.com/IBM-Blockchain/bcdb-server/internal/httputils"
 	"github.com/IBM-Blockchain/bcdb-server/pkg/constants"
 	"github.com/IBM-Blockchain/bcdb-server/pkg/cryptoservice"
 	"github.com/IBM-Blockchain/bcdb-server/pkg/types"
 	"github.com/gorilla/mux"
 )
 
-func MarshalOrPanic(response interface{}) []byte {
-	bytes, err := json.Marshal(response)
-	if err != nil {
-		panic(err)
-	}
-
-	return bytes
-}
-
-// SendHTTPResponse writes HTTP response back including HTTP code number and encode payload
-func SendHTTPResponse(w http.ResponseWriter, code int, payload interface{}) {
-	response, _ := json.Marshal(payload)
-	w.Header().Set("Content-Type", "application/json")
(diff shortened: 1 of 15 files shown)


internal/httphandler/utils.go
```diff
@@ -124,7 +80,7 @@ func extractVerifiedQueryPayload(w http.ResponseWriter, r *http.Request, queryTy
 	case constants.GetTxProof:
 		blockNum, txIndex, err := getBlockNumAndTxIndex(params)
```
gennadylaventman, line 81: If in previous cases you replaced getSomething(params) function by new httputils.GetSomething(params), this one should go there as well...

internal/httphandler/utils.go
```diff
@@ -136,15 +92,15 @@ func extractVerifiedQueryPayload(w http.ResponseWriter, r *http.Request, queryTy
 	case constants.GetDataProof:
 		blockNum, err := getBlockNum(params)
```
gennadylaventman, line 93: Same here

internal/httphandler/utils.go
```diff
@@ -164,13 +120,13 @@ func extractVerifiedQueryPayload(w http.ResponseWriter, r *http.Request, queryTy
 	case constants.GetHistoricalData:
 		version, err := getVersion(params)
```
gennadylaventman, line 121: And here. btw, `getVersion` used multiple times

internal/httputils/utils.go
```diff
@@ -0,0 +1,68 @@
...
+func JsonMarshalOrPanic(response interface{}) []byte {
+	bytes, err := json.Marshal(response)
+	if err != nil {
+		panic(err)
+	}
+
+	return bytes
+}
+
+// SendHTTPResponse writes HTTP response back including HTTP code number and encode payload
```
gennadylaventman, line 24: I don't think we encode whole payload anymore, so encode may go

internal/httphandler/utils.go
tock-ibm: I only moved to the common `httputils` package the minimal set of functions I needed for the catch-up service... (i.e. what's coming in the next PR). I think it is better to keep them here for now,, and if I'll need them there, I'll move them at that time. 
What do you think?

internal/httputils/utils.go
tock-ibm: Not sure I understand what you mean here?

gennadylaventman: In comment `encode payload` - we don't encode payloads anymore

internal/httphandler/utils.go
tock-ibm: None

internal/httphandler/utils.go
tock-ibm: Same as above, not needed by the catchup service.

tock-ibm: Done

internal/httphandler/utils.go
tock-ibm: Done

internal/httputils/utils.go
tock-ibm: But we do, using json.Marshal... mainly used for errors though... `payload` is the parameter name

gennadylaventman (approved):

PR #175, opened 2021-09-20
order preseving encoding for +ve and -ve numbers

cendhu:
When we use order preserving encoded bytes of number in the
index entry, the json marshaler converts the bytes to base64
encoding which does not preserve the order.

This commit use hex encoding to convert a given number into
an order preserving string.

### DIFF
--- a/internal/stateindex/encoding.go
+++ b/internal/stateindex/encoding.go
@@ -1,13 +1,16 @@
 package stateindex
 
 import (
+	"errors"
 	"math"
 )
 
-// EncodeOrderPreservingVarUint64 returns a byte-representation for a uint64 number such that
+const hextable = "0123456789abcdef"
+
+// EncodeOrderPreservingVarUint64 returns a string-representation for a uint64 number such that
 // all zero-bits starting bytes are trimmed in order to reduce the length of the array
 // For preserving the order in a default bytes-comparison, first byte contains the number of remaining bytes.
-func EncodeOrderPreservingVarUint64(n uint64) []byte {
+func EncodeOrderPreservingVarUint64(n uint64) string {
 	var bytePosition int
 	for bytePosition = 0; bytePosition <= 7; bytePosition++ {
 		if byte(n>>(56-(bytePosition*8))) != 0x00 {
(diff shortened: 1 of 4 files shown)


internal/stateindex/encoding.go
```diff
@@ -81,3 +98,35 @@ func decodeReverseOrderVarUint64(b []byte) uint64 {
 
 	return math.MaxUint64 - v
 }
+
+const hextable = "0123456789abcdef"
```
tock-ibm, line 102: `const` are usually at the top of the file, no?

internal/stateindex/encoding.go
```diff
@@ -81,3 +98,35 @@ func decodeReverseOrderVarUint64(b []byte) uint64 {
...
+// Encode implements hexadecimal encoding.
+func encode(dst, src []byte) int {
+	j := 0
+	for _, v := range src {
+		dst[j] = hextable[v>>4]
+		dst[j+1] = hextable[v&0x0f]
+		j += 2
+	}
+	return len(src) * 2
+}
```
tock-ibm, line 120: dead code

internal/stateindex/encoding.go
```diff
@@ -81,3 +98,35 @@ func decodeReverseOrderVarUint64(b []byte) uint64 {
...
+
+// fromHexChar converts a hex character into its value
+func fromHexChar(c byte) byte {
```
tock-ibm, line 123: as commented above, this may result in an error when the char is not `0-9` `a-f` `A-F`

internal/stateindex/encoding.go
```diff
@@ -16,24 +16,34 @@ func EncodeOrderPreservingVarUint64(n uint64) []byte {
 	}
 
 	size := int8(8 - bytePosition)
-	encodedBytes := make([]byte, size+1)
-	encodedBytes[0] = byte(size)
-	p := 1
+	encodedBytes := make([]byte, encodedLen(int(size)+1))
+	b := byte(size)
+	encodedBytes[0] = hextable[b>>4]
```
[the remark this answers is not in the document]
cendhu: Done.

internal/stateindex/encoding.go
```diff
@@ -16,24 +16,34 @@ func EncodeOrderPreservingVarUint64(n uint64) []byte {
...
 	}
-	return encodedBytes
+	return string(encodedBytes)
 }
 
 // decodeOrderPreservingVarUint64 decodes the number from the bytes obtained from method 'encodeOrderPreservingVarUint64'.
 // It returns the decoded number.
-func decodeOrderPreservingVarUint64(b []byte) uint64 {
-	size := int(b[0])
+func decodeOrderPreservingVarUint64(s string) uint64 {
```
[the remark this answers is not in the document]
cendhu: I was aware of this and kept it simple. Now, I am handing the error case too. Done.

internal/stateindex/encoding.go
```diff
@@ -42,7 +52,7 @@ func decodeOrderPreservingVarUint64(b []byte) uint64 {
 // are trimmed and replaced by the number of such trimmed bytes. This helps in reducing the size.
```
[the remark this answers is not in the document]
cendhu: Done

internal/stateindex/encoding.go
```diff
@@ -42,7 +52,7 @@ func decodeOrderPreservingVarUint64(b []byte) uint64 {
...
 // In the byte order comparison this encoding ensures that encodeReverseOrderVarUint64(A) > encodeReverseOrderVarUint64(B),
 // If B > A
```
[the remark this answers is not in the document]
cendhu: Done

internal/stateindex/encoding.go
```diff
@@ -42,7 +52,7 @@ func decodeOrderPreservingVarUint64(b []byte) uint64 {
...
-func EncodeReverseOrderVarUint64(n uint64) []byte {
+func EncodeReverseOrderVarUint64(n uint64) string {
 	n = math.MaxUint64 - n
```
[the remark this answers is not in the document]
cendhu: One this part is exactly the same
```
	j := 2
	for r := bytePosition; r <= 7; r++ {
		b = byte(n >> (56 - (r * 8)))
		encodedBytes[j] = hextable[b>>4]
		encodedBytes[j+1] = hextable[b&0x0f]
		j += 2
	}
	return string(encodedBytes)
```

Though I can use a common function, I don't prefer it here as duplication is not much. One uses `size` as a prefix and another as `bytePosition`. While I can make a common function, I felt that it break the readability and hides the difference (which I do not want).

internal/stateindex/encoding.go
cendhu: Done

internal/stateindex/encoding.go
cendhu: Done

internal/stateindex/encoding.go
cendhu: Yes, it was an intentional but changed as per your suggestion.

internal/stateindex/encoding.go
```diff
@@ -16,33 +19,51 @@ func EncodeOrderPreservingVarUint64(n uint64) []byte {
...
 
 	size := int8(8 - bytePosition)
-	encodedBytes := make([]byte, size+1)
-	encodedBytes[0] = byte(size)
-	p := 1
+	encodedBytes := make([]byte, encodedLen(int(size)+1))
+	b := byte(size)
+	// given that the size is <= 8, the 4 most significant bits would always be 0
+	encodedBytes[0] = '0'
+	encodedBytes[1] = hextable[b&0x0f]
```
tock-ibm, line 26: given that the size is <= 8, you don't need to do `b & 0x0f` as well...

internal/stateindex/encoding.go
```diff
@@ -53,31 +74,77 @@ func EncodeReverseOrderVarUint64(n uint64) []byte {
 	}
 
 	size := int8(8 - bytePosition)
-	encodedBytes := make([]byte, size+1)
-	encodedBytes[0] = byte(bytePosition)
-	p := 1
+	encodedBytes := make([]byte, encodedLen(int(size)+1))
+	b := byte(bytePosition)
+	encodedBytes[0] = '0'
+	encodedBytes[1] = hextable[b&0x0f]
```
tock-ibm, line 80: same here, no need to `& 0x0f`

internal/stateindex/encoding.go
```diff
@@ -53,31 +74,77 @@ func EncodeReverseOrderVarUint64(n uint64) []byte {
...
+		return 0, err
+	}
+
+	secondChar, err := fromHexChar(b[location+1])
+	if err != nil {
+		return 0, err
+	}
+
+	return (firstChar << 4) | secondChar, nil
+
```
tock-ibm, line 131: remove space

internal/stateindex/encoding.go
```diff
@@ -53,31 +74,77 @@ func EncodeReverseOrderVarUint64(n uint64) []byte {
...
+func fromHexChar(c byte) (byte, error) {
+	switch {
+	case '0' <= c && c <= '9':
+		return c - '0', nil
+	case 'a' <= c && c <= 'f':
+		return c - 'a' + 10, nil
+	case 'A' <= c && c <= 'F':
+		return c - 'A' + 10, nil
+	default:
+		return 0, errors.New("invalid character")
```
tock-ibm, line 148: how about `invalid hex character: z`?

cendhu: Done

internal/stateindex/encoding.go
cendhu: Done

tock-ibm (approved):

PR #179, opened 2021-09-22
Support numbers in query

cendhu:
This commit enables the usage of number within a query to do operation such as > -100 and < 50, etc..


internal/stateindex/encoding_test.go
```diff
@@ -7,6 +7,40 @@ import (
...
+			name: "minimum negative value",
+			n:    -1,
+		},
+		{
+			name: "maximum positive value",
+			n:    math.MaxInt64,
+		},
+		{
+			name: "minimum positive value",
+			n:    math.MaxInt64,
```
tock-ibm, line 29: min is 0, and add +1 as well

tock-ibm: I would also fill an array with a few more values just for good measure, like maybe +/- 3^k, k=1:39,
sort the array, encode, sort, decode, and compare

tock-ibm (commented):
Just minor testing suggestions, otherwise looks good

Issue #174, opened 2021-09-15
manage memory utilization of query executor


cendhu:
We need to control the total amount of memory used by the query executor to avoid OOM error. As each query collects keys and also values, the amount of memory used could go beyond the total RAM size and it could result in the panic of the bcdb node. 

Hence, we need to introduce memory limit for the query executor and ensure that the in-memory data does not go beyond this limit. When the limit reaches, we can store the partial results to the disk and control the memory usage as done in PostgreSQL. When we address the issue, we can add more design details here.

PR #181, opened 2021-09-23
Rest api for query

cendhu:


internal/httphandler/data_request_handler.go
```diff
@@ -144,3 +145,40 @@ func (d *dataRequestHandler) dataTransaction(response http.ResponseWriter, reque
 
 	d.txHandler.handleTransaction(response, txEnv, timeout)
 }
+
+func (d *dataRequestHandler) dataJSONQuery(response http.ResponseWriter, request *http.Request) {
+	payload, respondedErr := extractVerifiedQueryPayload(response, request, constants.PostDataQuery, d.sigVerifier)
+	if respondedErr {
+		return
```
C0rWin, line 152: Shouldn't you reply here back to the client with the HTTP error?

cendhu: Hi Artem, How are you doing?

`extractVerifiedQueryPayload()` has sent the error already and that's why it is returning `respondedErr`. This design has been in play since the beginning.

C0rWin: Doing good ;) 

> extractVerifiedQueryPayload() has sent the error already, and that's why it is returning respondedErr. This design has been in play since the beginning.

Too long haven't looked at the code and forgot it :)

PS. How are you doing?

internal/bcdb/worldstate_query_processor.go
```diff
@@ -140,3 +143,58 @@ func (q *worldstateQueryProcessor) getNodeConfig(nodeID string) (*types.GetNodeC
...
+		// TODO: we can store the ACL as value in the indexEntry. With that, we can avoid reading the whole value
+		// to perform the access control - issue #152
+		acl := metadata.GetAccessControl()
+		if acl != nil {
+			if !acl.ReadUsers[querierUserID] && !acl.ReadWriteUsers[querierUserID] {
+				continue
+			}
+		}
+
+		results = append(
```
tock-ibm, line 187: TODO 
- about controlling the returned number of keys, paging, etc... 
- controlling the size in bytes
- skipping, etc

internal/bcdb/worldstate_query_processor.go
```diff
@@ -140,3 +143,58 @@ func (q *worldstateQueryProcessor) getNodeConfig(nodeID string) (*types.GetNodeC
 
 	return c, nil
 }
+
+func (q *worldstateQueryProcessor) executeJSONQuery(dbName, querierUserID string, query []byte) (*types.DataQueryResponse, error) {
```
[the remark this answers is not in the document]
cendhu: > We'll have to add support for that later, and it would be good not to change the API when that happens, just expand it.

which API? the SDK API exported to the user can support limit and skip now also. When we support `limit` and `skip`, we need to change the SDK. I prefer to do it incrementally. 

design have kept all these in mind. Code is easily extendable.

internal/httphandler/utils.go
```diff
@@ -195,6 +196,28 @@ func extractVerifiedQueryPayload(w http.ResponseWriter, r *http.Request, queryTy
 			Id:      params["id"],
 			Version: version,
 		}
+	case constants.PostDataQuery:
+		if r.Body == nil {
+			httputils.SendHTTPResponse(w, http.StatusBadRequest, &types.HttpResponseErr{ErrMsg: "query is empty"})
+			return nil, true
+		}
+
+		b, err := io.ReadAll(r.Body)
```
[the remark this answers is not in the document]
cendhu: how does it passes the CI then? 

This is because the CI uses 1.16 since we moved to GitHub. I remember updating my local version. I have updated the go.mod to 1.16. We should in fact bump it to 1.17 soon

pkg/constants/http.go
```diff
@@ -59,6 +60,19 @@ func URLForGetData(dbName, key string) string {
...
+	return path.Join(DataEndpoint, dbName)
+}
+
+// URLForJSONQuery returns url for GET request to retrieve
+// key-value pairs present in the dbName which are matching the
+// given JSON query criteria
+func URLForJSONQuery(dbName string) string {
+	return DataEndpoint + path.Join("jsonquery", dbName)
+}
+
```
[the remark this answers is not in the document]
cendhu: Good catch. 

Yes, initially, I had introduce a separate path for the query. Later, I decided to reuse the same `data/dbname/`. I forgot to remove this. Now that we are using `data/dbname/jsonquery`, we can reuse this with minor mods.

internal/bcdb/worldstate_query_processor.go
cendhu: skipping is a feature. Not a todo and it requires a story. For others, we have a story in terms of restricting the response size (https://github.com/IBM-Blockchain/bcdb-server/issues/182) and memory management (https://github.com/IBM-Blockchain/bcdb-server/issues/174).  As I am working on those currently, I do not see value in putting a TODO now.

internal/httphandler/data_request_handler.go
```diff
@@ -144,3 +145,40 @@ func (d *dataRequestHandler) dataTransaction(response http.ResponseWriter, reque
...
+	}
+	query := payload.(*types.DataJSONQuery)
+
+	if !d.db.IsDBExists(query.DbName) {
+		httputils.SendHTTPResponse(response, http.StatusBadRequest, &types.HttpResponseErr{
+			ErrMsg: "'" + query.DbName + "' does not exist",
+		})
```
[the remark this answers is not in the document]
cendhu: For now, this is consistent with the existing behaviour. hence, I am leaving it as-is.

internal/bcdb/worldstate_query_processor.go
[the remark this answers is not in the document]
cendhu: Done.

cendhu:
Addressed most of the comments. Ready for the review

internal/bcdb/db.go
```diff
@@ -395,6 +418,45 @@ func (d *db) GetData(dbName, querierUserID, key string) (*types.GetDataResponseE
...
+// 			},
+// 			"attr2": {          -- a field in the json document
+// 				"$eq": true     -- value criteria for the field
+// 			},
+// 			"attr3": {          -- a field in the json document
+// 				"$lt": "a2"     -- a field in the json document
+// 			}
+// 		}
+//   }
+// }
```
tock-ibm, line 441: I don't think you need to put the same comments twice, either on the interface or here is fine, I think

internal/bcdb/worldstate_query_processor.go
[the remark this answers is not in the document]
tock-ibm: Yes, I meant that the structure will be at least:
```
{
  "selector" : {...}
}
```
The rest could follow later.

internal/bcdb/worldstate_query_processor.go
tock-ibm: I agree, by TODO I meant have a story, not necessarily a code comment. Just using the opportunity to flash out issues so we won't forget.

internal/bcdb/worldstate_query_processor_test.go
```diff
@@ -324,8 +321,296 @@ func TestGetData(t *testing.T) {
...
+							"$gt": "",
+							"$lte": "d"
+						}
+					}
+				}`,
+			),
+			expectedErr: "no user can directly read from a system database [" + worldstate.ConfigDBName + "]",
+		},
+		{
+			name:   "user does not have read permission",
```
tock-ibm, line 544: another test case: what happens if the query contains an attribute that is not indexed?

internal/httphandler/data_request_handler.go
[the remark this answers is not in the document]
tock-ibm: what existing behavior? maybe open an issue/bug report to change all such behavior, because it is wrong.

internal/httphandler/utils.go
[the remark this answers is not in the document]
tock-ibm: I will update mine as well

pkg/constants/http.go
```diff
@@ -59,6 +60,19 @@ func URLForGetData(dbName, key string) string {
...
+// key-value pairs matching a given query criteria
+func URLForDataQuery(dbName string) string {
+	return path.Join(DataEndpoint, dbName)
+}
+
+// URLForJSONQuery returns url for GET request to retrieve
+// key-value pairs present in the dbName which are matching the
+// given JSON query criteria
+func URLForJSONQuery(dbName string) string {
+	return DataEndpoint + path.Join(dbName, "jsonquery")
```
tock-ibm, line 73: `return path.Join(DataEndpoint, dbName, "jsonquery")`

pkg/constants/http.go
```diff
@@ -59,6 +60,19 @@ func URLForGetData(dbName, key string) string {
 	return DataEndpoint + path.Join(dbName, key)
 }
 
+// URLForDataQuery returns url for POST request to retrieve
+// key-value pairs matching a given query criteria
+func URLForDataQuery(dbName string) string {
+	return path.Join(DataEndpoint, dbName)
+}
```
tock-ibm, line 67: why do we need this? I don't see this path ever being used at the moment

pkg/constants/http_test.go
```diff
@@ -24,6 +24,20 @@ func TestURLConstruction(t *testing.T) {
 			},
 			expectedURL: "/data/db1/key1",
 		},
+		{
+			name: "ExecuteQuery",
+			execute: func() string {
+				return URLForDataQuery("db1")
+			},
+			expectedURL: "/data/db1",
+		},
```
tock-ibm, line 33: why do we need this? I don't see this path ever being used at the moment

pkg/constants/query.go
```diff
@@ -14,4 +14,7 @@ const (
 	QueryOpLesserThan         = "$lt"
 	QueryOpGreaterThanOrEqual = "$gte"
 	QueryOpLesserThanOrEqual  = "$lte"
+
+	// Top-level fields allowed in the query
+	Selector = "selector"
```
tock-ibm, line 19: Prefix with `QueryField*` to avoid confusion, same logic as having `QueryOp*` above.
I expect we'll have more fields in the future, like  `QueryFieldLimit` etc.

protos/query.proto
```diff
@@ -202,3 +202,9 @@ message GetMostRecentUserOrNodeQuery {
     string id = 3;
     Version  version = 4;
 }
+
+message DataJSONQuery {
+    string user_id = 1;
+    string db_name = 2;
+    string query = 3;
```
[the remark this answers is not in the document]
tock-ibm: OK, I agree, the server will enforce an internal structure in the query string, like looking for the "selector" field. The SDK will support in populating that json string in a way the server expects it.

internal/bcdb/worldstate_query_processor_test.go
cendhu: Why just stop there? why not test the other package again here? When the attribute is not indexed, the queryexecutor would return an error.

pkg/constants/query.go
cendhu: Done

pkg/constants/http_test.go
cendhu: Done

internal/httphandler/data_request_handler.go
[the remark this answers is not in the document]
cendhu: Yes, it is kind of not found 404. 

I also saw that couchdb returns database does not exist. 

https://groups.google.com/g/couchdb-user-archive/c/zUu76O7A534?pli=1

PR #185, opened 2021-09-25
Use leveldb snapshot to process queries

cendhu:
as query processing can take time, we cannot hold
the read lock on the database for that time. If we
do that, the commit process would get impacted and
hence, the performance.

Instead of locking the database, we use leveldb
snapshots to process the queries. As a result, the
query might return data that might have been deleted
by the lastest block. However, this behavior is okay
because we have a similar scenario in the lock based
system too, i.e., the returned results could be deleted
by blocks just after the submitter receives the results.


internal/worldstate/leveldb/snapshot.go
```diff
@@ -0,0 +1,113 @@
...
+	if !ok {
+		return nil, nil, errors.New(dbName + " is needed to fetch the index definiton and is not snapshotted")
+	}
+
+	dbval, err := lSnap.Get([]byte(key), &opt.ReadOptions{})
+	if err == leveldb.ErrNotFound {
+		return nil, nil, nil
+	}
+	if err != nil {
+		return nil, nil, errors.WithMessagef(err, "failed to retrieve leveldb key [%s] from database %s", key, dbName)
```
gennadylaventman, line 64: "failed to retrieve leveldb key [%s] from database %s snapshot"

internal/worldstate/db.go
```diff
@@ -56,6 +56,11 @@ type DB interface {
 	// the caller wants from the first key in the database (lexicographic order). An empty
 	// endKey (i.e., "") denotes that the caller wants till the last key in the database (lexicographic order).
 	GetIterator(dbName string, startKey, endKey string) (Iterator, error)
+	// GetSnapshot returns a latest snapshot of the given DB along with all system databases.
```
gennadylaventman, line 59: `// GetSnapshot returns a latest snapshot of the given DB along with all system databases`???
From code no snapshot for all system databases created

internal/bcdb/worldstate_query_processor.go
```diff
@@ -162,15 +161,27 @@ func (q *worldstateQueryProcessor) executeJSONQuery(dbName, querierUserID string
 		}
 	}
 
-	keys, err := q.jsonQueryExecutor.ExecuteQuery(dbName, query)
+	snapshots, err := q.db.GetSnapshot(
```
gennadylaventman, line 164: nit - why result of `GetSnapshot` assign to variable `snapshots`

internal/bcdb/worldstate_query_processor.go
```diff
@@ -162,15 +161,27 @@ func (q *worldstateQueryProcessor) executeJSONQuery(dbName, querierUserID string
...
+			worldstate.DatabasesDBName,
+			dbName,
+			stateindex.IndexDB(dbName),
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	jsonQueryExecutor := queryexecutor.NewWorldStateJSONQueryExecutor(snapshots, q.logger)
```
gennadylaventman, line 175: I didn't find the place where snapshot is released.

internal/worldstate/leveldb/commit_and_query.go
```diff
@@ -306,7 +306,8 @@ func (l *LevelDB) create(dbName string) error {
 	if err != nil {
 		return errors.WithMessagef(err, "failed to open leveldb file for database %s", dbName)
 	}
-
+	s, err := file.GetSnapshot()
+	s.Release()
```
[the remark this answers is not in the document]
cendhu: Good catch. I added this code in a random place to see the API for getting the snapshot from the level DB but forgot to remove these two lines of code :-)

internal/worldstate/leveldb/snapshot.go
cendhu: Done

internal/worldstate/db.go
cendhu: Done. This was the old design but later I changed it to []string so that the consumer can tell which system databases and user databases are needed.

internal/bcdb/worldstate_query_processor.go
cendhu: Done. I have changed the signature to `GetDBsSnapshot` and `DBsSnapshot` in leveldb package.

internal/bcdb/worldstate_query_processor.go
cendhu: Good catch. Done.

gennadylaventman (approved):

PR #186, opened 2021-09-27
Handle not equal to ($neq) in query

cendhu:
This commit enables the usage of $neq, i.e., not equal to,
conditions on all attributes types. Note that both `$eq` and
`$neq` cannot appear for an attribute (as it does not make sense).


internal/queryexecutor/range_query_planner.go
```diff
@@ -1,28 +1,43 @@
...
-	"errors"
-
 	"github.com/IBM-Blockchain/bcdb-server/internal/stateindex"
 	"github.com/IBM-Blockchain/bcdb-server/pkg/constants"
 )
 
 type rangeQueryPlan struct {
 	startKey *stateindex.IndexEntry
 	endKey   *stateindex.IndexEntry
+	seekKey  *stateindex.IndexEntry
```
tock-ibm, line 11: I would call that `excludeKeys` and make it an array, to allow multiple $neq conditions in a range

internal/queryexecutor/range_query_planner.go
```diff
@@ -1,28 +1,43 @@
...
+			continue
 		}
+
+		seekKey = &stateindex.IndexEntry{
+			Attribute:     attribute,
+			Type:          conds.valueType,
+			ValuePosition: stateindex.Existing,
+			Value:         v,
+			KeyPosition:   stateindex.Ending,
+		}
```
tock-ibm, line 37: can we accumulate the $neq values to an array?

internal/queryexecutor/execute_conditions.go
```diff
@@ -151,6 +153,32 @@ func (e *WorldStateJSONQueryExecutor) execute(dbName string, attribute string, c
...
+		seekKey, err := plan.seekKey.String()
+		if err != nil {
+			return nil, err
+		}
+		itemExist := iter.Seek([]byte(seekKey))
+		seeked = true
+
+		if !itemExist {
+			break
+		}
```
tock-ibm, line 175: Do we expect this to ever match? keys in the index have always Existing/`v`/Existing, so if i understand correctly, since we are seeking Existing/`v`/Ending, this will get the the next value after `v`, no?
If so, why not just continue to iterate?

internal/queryexecutor/range_query_planner.go
cendhu: Done

internal/queryexecutor/range_query_planner.go
```diff
@@ -36,15 +71,24 @@ func createQueryPlan(attribute string, conds *attributeTypeAndConditions) (*rang
...
 			setPlanForSingleCondition(c, v, p)
-			return p, nil
 		}
+	} else {
+		setPlanForMultipleConditions(conds, p)
 	}
 
-	setPlanForMultipleConditions(conds, p)
-
+	p.excludeKeys = excludeKeys
```
tock-ibm, line 91: As an optimization, after setting the start and end key, you could prune the excludeKeys for values out side that range?
This does not affect correctness as the executor will never match the value of entries from within the range to an exlude-key from out side it... but, it might save some time. 

In addition, I am not sure I've seen a test for a range [x,y] with exclude keys outside it?

tock-ibm: Will be addressed in one of the following PRs.

tock-ibm (approved):

PR #191, opened 2021-10-06
use context to avoid unnecessary work in query execution 

cendhu:
When the following occurs, we might do unnecessary query processing at node:
  -- client request timeout
  -- connection between the client and server drops
  -- server shutdown

By acquiring a context from the http request, we avoid unnecessary work in query processing for all the above three cases.


gennadylaventman (commented):
One comment

PR #193, opened 2021-10-06
Limit state transfer response size

tock-ibm:
Limit the server's state transfer response size, in order to protect the server
from client requests that return so many blocks as to exhaust the server's memory.
This also protects the client from receiving a response too big to handle.


internal/comm/httptransport.go
```diff
@@ -63,7 +63,7 @@ func NewHTTPTransport(config *Config) *HTTPTransport {
 		logger:         config.Logger,
 		localConf:      config.LocalConf,
 		catchUpClient:  NewCatchUpClient(config.Logger),
-		catchupHandler: NewCatchupHandler(config.Logger, config.LedgerReader),
+		catchupHandler: NewCatchupHandler(config.Logger, config.LedgerReader, 0), //TODO make max-response-bytes configurable
```
cendhu, line 66: I assume it to be part of the local config file. I am also fine if we fix it to 100 MB always and keep the config file with a smaller number of important parameters.

internal/comm/catchup_handler_test.go
```diff
@@ -247,3 +247,110 @@ func TestCatchupHandler_ServeHTTP_Blocks(t *testing.T) {
...
+		if n==1 {
+			b1Size = len(httputils.MarshalOrPanic(block))
+		}
+		if n<6 {
+			b5Size += len(httputils.MarshalOrPanic(block))
+		}
+	}
+
+	t.Run("too many blocks in request", func(t *testing.T) {
+		h := comm.NewCatchupHandler(lg, ledger1, b5Size+1) // 5 blocks in response
```
cendhu, line 282: Is there any reason to do `b5Size+1` instead of `b5Size`? I assume none and both should work.

internal/comm/catchup_handler_test.go
```diff
@@ -247,3 +247,110 @@ func TestCatchupHandler_ServeHTTP_Blocks(t *testing.T) {
...
+		h.ServeHTTP(resp, req)
+		require.Equal(t, http.StatusOK, resp.Result().StatusCode)
+
+		_, params, err := mime.ParseMediaType(resp.Result().Header.Get("Content-Type"))
+		require.NoError(t, err)
+		boundary, ok := params["boundary"]
+		require.True(t, ok)
+		mr := multipart.NewReader(resp.Result().Body, boundary)
+		bNum := uint64(2)
+		for part, errP := mr.NextPart(); errP == nil; part, errP = mr.NextPart() {
```
cendhu, line 302: we can have `for { if errP == io.EOF {break}}` for better readability. We also need to ensure that no err other than `io.EOF` is returned. The same is applicable for the other sub-test with a smaller maxSize.

cendhu (commented):
LGTM. A few minor comments.

internal/comm/catchup_handler_test.go
tock-ibm: Done

PR #194, opened 2021-10-08
maintain order between -ve and +ve number

cendhu:
This commits make the -ve number appear first
in the order before the +ve number. Further,
the value order is preserved too (i.e., -1000
appear before -100 and -90 appears before 10,
and so on)

### DIFF
--- a/internal/stateindex/encoding.go
+++ b/internal/stateindex/encoding.go
@@ -5,11 +5,16 @@ import (
 	"math"
 )
 
-const hextable = "0123456789abcdef"
+const (
+	hextable     = "0123456789abcdef"
+	reverseOrder = '0'
+	normalOrder  = '1'
+)
 
 // EncodeOrderPreservingVarUint64 returns a string-representation for a uint64 number such that
 // all zero-bits starting bytes are trimmed in order to reduce the length of the array
-// For preserving the order in a default bytes-comparison, first byte contains the number of remaining bytes.
+// For preserving the order in a default bytes-comparison, first byte contains the type of
+// encoding and the second byte contains the number of remaining bytes.
 func EncodeOrderPreservingVarUint64(n uint64) string {
 	var bytePosition int
 	for bytePosition = 0; bytePosition <= 7; bytePosition++ {
@@ -21,8 +26,9 @@ func EncodeOrderPreservingVarUint64(n uint64) string {
 	size := int8(8 - bytePosition)
 	encodedBytes := make([]byte, encodedLen(int(size)+1))
 	b := byte(size)
-	// given that the size is <= 8, the 4 most significant bits would always be 0
-	encodedBytes[0] = '0'
+	// given that size will never be greater than 8, we use the first
+	// byte to denote the normal order encoding
+	encodedBytes[0] = normalOrder
 	encodedBytes[1] = hextable[b]
 
 	j := 2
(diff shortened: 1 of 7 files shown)


tock-ibm:
There is a problem with this approach:
```
func EncodeReverseOrderVarUint64(n uint64) string {
	n = math.MaxUint64 - n
        ...
}
```
means that small negative numbers  `-1` are represented as big numbers `math.MaxUint64-1`.
This kind of defeats the purpose of var-int, that assumes small (absolute) numbers are much more prevalent than large ones.

The approach represented by [this article](https://medium.com/@neunhoef/sorting-number-strings-numerically-335863473b76) does not have that. Instead, when you encode the digits of a negative number, you flip the index into the hex table. So, for positive number represented as an hex without leading zeroes (e.g. `6e` = `110` ):

Nd | hex-digits

where Nd is the hex-digit that represents the number of hex-digits. For an int64, it is 0-8.

Thus:
encode( `6e`) = `26e`
encode( `46e`) = `346e`

For a negative number in hex format (e.g `-6e` = `-110`)

'-' | flip(Nd) | flip(hex-digits)

where flip(hex-digit) is just a mapping
 0 -> f 
 1 -> e 
 2 -> d
 3 -> c
 4 -> b
 5 -> a
 6 -> 9
 7 -> 8
 8 -> 7
 ... etc, 
i.e. looking into the `hexTable` with `15-index`. 
Note that '-' is smaller than the smallest Nd for positive numbers, '0'. Thus:

* encode( `-6e` ) = `-d91`
* encode( `-5e` ) = `-da1`

tock-ibm:
I think it would be better to have single exported method 
```
  func EncodeVarInt64(n int64) string
```
and test that directly, rather than switch between `EncodeReverseOrderVarUint64` and `EncodeOrderVarUint64` according to the sign of the incoming `int64`.

tock-ibm:
In addition, I think in 
```
enum IndexAttributeType {
  NUMBER = 0;
  STRING = 1;
  BOOLEAN = 2;
}
```
should be changed `NUMBER` to `INT64`. This way, we can later easily introduce DOUBLE (or FLOAT), as [described here](https://medium.com/@aravindet/order-preserving-encoding-for-floats-cde09c978629).

internal/stateindex/encoding.go
```diff
@@ -5,11 +5,16 @@ import (
 	"math"
 )
 
-const hextable = "0123456789abcdef"
+const (
+	hextable     = "0123456789abcdef"
+	reverseOrder = '0'
+	normalOrder  = '1'
```
tock-ibm, line 11: See comment https://github.com/IBM-Blockchain/bcdb-server/pull/194#issuecomment-939435457

internal/stateindex/encoding.go
```diff
@@ -5,11 +5,16 @@ import (
...
+)
 
 // EncodeOrderPreservingVarUint64 returns a string-representation for a uint64 number such that
 // all zero-bits starting bytes are trimmed in order to reduce the length of the array
-// For preserving the order in a default bytes-comparison, first byte contains the number of remaining bytes.
+// For preserving the order in a default bytes-comparison, first byte contains the type of
+// encoding and the second byte contains the number of remaining bytes.
 func EncodeOrderPreservingVarUint64(n uint64) string {
```
tock-ibm, line 18: See https://github.com/IBM-Blockchain/bcdb-server/pull/194#issuecomment-939435482

cendhu:
> There is a problem with this approach:
> 
> ```
> func EncodeReverseOrderVarUint64(n uint64) string {
> 	n = math.MaxUint64 - n
>         ...
> }
> ```
> 
> means that small negative numbers `-1` are represented as big numbers `math.MaxUint64-1`. This kind of defeats the purpose of var-int, that assumes small (absolute) numbers are much more prevalent than large ones.
> 

We have used this mechanism in Fabric. Hence, I prefer to use it. The word `VarInt` can be removed as we are anyway using extra character for each non-empty byte. Hence, I am not going to change this encoding scheme.

cendhu:
> In addition, I think in
> 
> ```
> enum IndexAttributeType {
>   NUMBER = 0;
>   STRING = 1;
>   BOOLEAN = 2;
> }
> ```
> 
> `NUMBER` should be changed to `INT64`. This way, we can later easily introduce DOUBLE (or FLOAT), as [described here](https://medium.com/@aravindet/order-preserving-encoding-for-floats-cde09c978629).



> In addition, I think in
> 
> ```
> enum IndexAttributeType {
>   NUMBER = 0;
>   STRING = 1;
>   BOOLEAN = 2;
> }
> ```
> 
> `NUMBER` should be changed to `INT64`. This way, we can later easily introduce DOUBLE (or FLOAT), as [described here](https://medium.com/@aravindet/order-preserving-encoding-for-floats-cde09c978629).

Unrelated to this PR. Will be fixed separately.

internal/stateindex/index_entries.go
```diff
@@ -229,7 +224,7 @@ func partialIndexEntriesForValue(v reflect.Value, index map[string]types.IndexAt
 					ValuePosition: Existing,
 					KeyPosition:   Existing,
 				}
-				e.Metadata, e.Value = GetMetadataAndValue(value, valueType)
+				e.Value = GetMetadataAndValue(value, valueType)
```
tock-ibm, line 227: `GetMetadataAndValue` There is no Metadata anymore...

PR #205, opened 2021-10-17
Update docker build

tock-ibm:

### DIFF
--- a/docs/curl/build.md
+++ b/docs/curl/build.md
@@ -16,10 +16,10 @@ the blockchain DB will be deployed.
 
 To build the executable binary, the following steps need to be executed
 
-  1. To clone the repository, first, create required directories using the command `mkdir -p github.com/IBM-Blockchain`
-  2. Change the current working directory to the above created folder by issing the command `cd github.com/IBM-Blockchain`
-  3. Clone this repository with `git clone https://github.com/IBM-Blockchain/bcdb-server`
-  4. Change the current working directory to the repository root directory by issuing `cd bcdb-server`
+  1. To clone the repository, first, create required directories using the command `mkdir -p github.com/hyperledger-labs`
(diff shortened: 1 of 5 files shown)


docs/curl/build.md
```diff
@@ -107,7 +107,7 @@ make docker
 #### Start docker container
 To invoke bcdb docker container, you can run:
```
gennadylaventman, line 108: bcdb -> orion

images/Dockerfile
```diff
@@ -15,16 +15,16 @@ RUN apk add --no-cache \
...
-FROM bcdb-base
-VOLUME /etc/bcdb-server
-VOLUME /var/bcdb-server
-COPY --from=bcdb /go/src/github.com/IBM-Blockchain/bcdb-server/bin/bdb /usr/local/bin
-COPY --from=bcdb /go/src/github.com/IBM-Blockchain/bcdb-server/sampleconfig/ /etc/bcdb-server
+FROM orion-base
+VOLUME /etc/orion-server
+VOLUME /var/orion-server
+COPY --from=orion-server /go/src/github.com/hyperledger-labs/orion-server/bin/bdb /usr/local/bin
+COPY --from=orion-server /go/src/github.com/hyperledger-labs/orion-server/sampleconfig/ /etc/bcdb-server
```
gennadylaventman, line 28: bcdb-server -> orion-server. It doesn't matter too much, because volume configuration in `docker run` will override it in any case, but it rare case user runs docker without volume mapping, it will cause some problems.

gennadylaventman (commented):
Two small comments

images/Dockerfile
tock-ibm: good catch

gennadylaventman (approved):

PR #206, opened 2021-10-17
Release txs in block-creator

tock-ibm:
When the replicator rejects a block, release TXs in the block from the PendingTxs queue.

### DIFF
--- a/internal/blockcreator/blockcreator.go
+++ b/internal/blockcreator/blockcreator.go
@@ -5,6 +5,7 @@ package blockcreator
 import (
 	"github.com/hyperledger-labs/orion-server/internal/blockstore"
 	ierrors "github.com/hyperledger-labs/orion-server/internal/errors"
+	"github.com/hyperledger-labs/orion-server/internal/httputils"
 	"github.com/hyperledger-labs/orion-server/internal/queue"
 	"github.com/hyperledger-labs/orion-server/pkg/logger"
 	"github.com/hyperledger-labs/orion-server/pkg/types"
@@ -107,7 +108,7 @@ func (b *BlockCreator) Start() {
 				continue
 			}
 
-			blkNum := b.nextBlockNumber
+			blkNum := b.nextBlockNumber //TODO move block numbering to replication, see: https://github.com/hyperledger-labs/orion-server/issues/200
 
 			baseHeader, err := b.createBaseHeader(blkNum)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


internal/blockcreator/blockcreator.go
```diff
@@ -148,13 +149,17 @@ func (b *BlockCreator) Start() {
 				// This may happen when shutting down the server. 'continue' will eventually pick up the stop signal.
 				b.logger.Warnf("block submission to block-replicator failed, dropping block, shutting down, because: %s", err)
 				continue
+
 			case *ierrors.NotLeaderError:
 				b.logger.Warnf("block submission to block-replicator failed, because this node is not a leader: %s", err)
-				//TODO reject or redirect all TXs in the block via the pending-tx component.
-				// If there is another leader then redirect, else reject.
+				// Releasing with an error will reject or redirect all sync TXs in the block via the pending-tx component.
+				// If there is another leader it will redirect, else reject. Async TXs will be removed.
```
cendhu, line 156: How does the redirection work? Is it done automatically by the raft or we need to implement it?

internal/blockcreator/blockcreator_test.go
```diff
@@ -3,12 +3,17 @@
 package blockcreator_test
 
 import (
+	"fmt"
+	ierrors "github.com/hyperledger-labs/orion-server/internal/errors"
```
cendhu, line 7: fix imports

internal/httputils/blocks.go
```diff
@@ -0,0 +1,72 @@
+// Copyright IBM Corp. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package httputils
+
+//TODO rename package to utils, and utils.go to http.go
```
cendhu, line 6: I assume there is an issue with this.

cendhu (commented):
Just one question before merging.

internal/blockcreator/blockcreator.go
tock-ibm: When a tx promise is waited upon by a request go-routine, and it is release with an error, the handler will process the error; 
- if it is a NotLeader with an alternative leader, it will with redirect (307) the client to the new leader;
- if their is no leader it will reject with a "service unavailable"  (503) 

This is already in place, see:
https://github.com/hyperledger-labs/orion-server/blob/04f19b212c8428d927c214babe390258e540307b/internal/httphandler/tx_handler.go#L30

async requests are dropped at this stage, as we don't forward internally now.

internal/blockcreator/blockcreator_test.go
```diff
@@ -3,34 +3,40 @@
...
 	"github.com/hyperledger-labs/orion-server/internal/blockcreator/mocks"
 	"github.com/hyperledger-labs/orion-server/internal/blockstore"
+	ierrors "github.com/hyperledger-labs/orion-server/internal/errors"
 	"github.com/hyperledger-labs/orion-server/internal/queue"
 	"github.com/hyperledger-labs/orion-server/internal/worldstate"
 	"github.com/hyperledger-labs/orion-server/internal/worldstate/leveldb"
 	"github.com/hyperledger-labs/orion-server/pkg/logger"
 	"github.com/hyperledger-labs/orion-server/pkg/types"
-	"github.com/golang/protobuf/proto"
+	"github.com/stretchr/testify/assert"
```
cendhu, line 24: remove this import in the next PR

cendhu (approved):

PR #217, opened 2021-10-21
Block numbering in replicator

tock-ibm:
The block replicator is made in charge of numbering the blocks and setting the last header base hash.
The block replicator keeps track of the last committed block, and the last proposed block, and
updates them whe leadership is assumed and lost.

The block numbers given the block creator are provisional, and number the sequence of proposals since
the node started its current incarnation.

### DIFF
--- a/internal/blockcreator/blockcreator.go
+++ b/internal/blockcreator/blockcreator.go
@@ -9,7 +9,6 @@ import (
 	"github.com/hyperledger-labs/orion-server/internal/queue"
 	"github.com/hyperledger-labs/orion-server/pkg/logger"
 	"github.com/hyperledger-labs/orion-server/pkg/types"
-	"github.com/pkg/errors"
 )
 
 //go:generate counterfeiter -o mocks/replicator.go --fake-name Replicator . Replicator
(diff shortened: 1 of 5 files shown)


internal/blockcreator/blockcreator.go
```diff
@@ -166,12 +152,7 @@ func (b *BlockCreator) Start() {
 				b.logger.Panicf("block submission to block-replicator failed: %v", err)
 			}
 
-			h, err := blockstore.ComputeBlockBaseHash(block)
-			if err != nil {
-				b.logger.Panicf("Error calculating block hash {%v}", err)
-			}
-			b.previousBlockHeaderBaseHash = h
-			b.nextBlockNumber++
+			b.nextProposalNumber++
```
cendhu, line 155: Earlier, block creator helped to parallelise certain tasks such as computing block hash, creating the base header by reading from the block store (IO). Now, this would be done serially within the replicator for each block. If this can be addressed within the replicator, it would be nice.

cendhu (commented):
Rebase

internal/blockcreator/blockcreator.go
tock-ibm: The go-routine that does that - the propose-loop go routine - is really lightly loaded, so this does not pose a performance hit in my opinion. More over, since the event-loop (which commits) and the propose-loop now communicate through memory to coordinate what was the last committed block, an access to IO is saved.

Note that the earlier version had a bug when a node was leader, lost leadership, and assumed leadership again.... the block creator never was sync'ed with the last committed block. This is the main reason behind this PR.

cendhu (approved):

PR #222, opened 2021-10-27
Limit the number of in flight blocks

tock-ibm:


internal/replication/blockreplicator.go
```diff
@@ -719,6 +746,10 @@ func (br *BlockReplicator) setLastCommittedBlock(block *types.Block) {
 	defer br.mutex.Unlock()
 
 	br.lastCommittedBlock = block
+	if br.numInFlightBlocks > 0 { // only reduce on the leader
```
cendhu, line 749: this check is nothing but a `AmIALeader()` check?

internal/replication/blockreplicator.go
```diff
@@ -312,6 +315,13 @@ Event_Loop:
 
 		case <-br.stopCh:
 			br.lg.Info("Stopping block replicator")
+
+			// notify the propose-loop go-routine in case it is waiting for blocks to commit
+			br.mutex.Lock()
+			br.numInFlightBlocks = 0
+			br.condTooManyInFlightBlocks.Broadcast()
```
cendhu, line 322: Should this be placed within the lock?

internal/replication/blockreplicator.go
```diff
@@ -355,6 +365,8 @@ func (br *BlockReplicator) processLeaderChanges(leader uint64) {
 				br.lg.Panicf("Error computing base header hash of last commited block: %+v; error: %s",
 					br.lastCommittedBlock.GetHeader(), err)
 			}
+			br.numInFlightBlocks = 0
```
cendhu, line 368: why no lock for this? The same is applicable for the other locations too. Use atomic type instead.

internal/replication/blockreplicator.go
```diff
@@ -638,6 +650,21 @@ func (br *BlockReplicator) updateLastProposal(lastBlockProposed *types.Block) {
 		} else {
 			br.lg.Panicf("Failed to compute last block base hash: %s", err)
 		}
+		br.numInFlightBlocks++
+
+		if br.numInFlightBlocks > br.clusterConfig.ConsensusConfig.RaftConfig.MaxInflightBlocks {
+			br.lg.Infof("Number of in-flight blocks %d exceeds max %d, waiting for blocks to commit",
+				br.numInFlightBlocks, br.clusterConfig.ConsensusConfig.RaftConfig.MaxInflightBlocks)
+
+			for br.numInFlightBlocks > br.clusterConfig.ConsensusConfig.RaftConfig.MaxInflightBlocks {
```
cendhu, line 659: I think the `numInFlightBlocks` need to be an atomic variable so that we can load and set it atomically.

cendhu (commented):
Is there a way to test this limit behaviour?

internal/replication/blockreplicator.go
tock-ibm: `isLeader && num>0`, for double safety, yes

internal/replication/blockreplicator.go
tock-ibm: `br.numInFlightBlocks = 0` must be in the lock as it accessed by a different go-routine, and `br.condTooManyInFlightBlocks.Broadcast()` can be placed either within or without (as the documentation on Signal and Broadcast explains), the result will be (almost) the same. It will wake up the waiting routine, which will try and get the lock as it exits from Wait, will block for it, and when the lock is released it will get it and discover the value change. The same can be done with a Broadcast after the Unlock(), and the waking go routine will also try to lock as it exits Wait. I prefer the former, there is no difference in safety. See:
https://go101.org/article/concurrent-synchronization-more.html

internal/replication/blockreplicator.go
tock-ibm: This is in a method that starts with 
```
func (br *BlockReplicator) processLeaderChanges(leader uint64) {
	br.mutex.Lock()
	defer br.mutex.Unlock()
```

internal/replication/blockreplicator.go
tock-ibm: This is in a method 
```
// updateLastProposal updates the last block proposed in order to keep track of block numbering.
func (br *BlockReplicator) updateLastProposal(lastBlockProposed *types.Block) {
	br.mutex.Lock()
	defer br.mutex.Unlock()
```

In general, `numInFlightBlocks` must be changed with a mutex locked for the waiting go-routine (event-loop) to see the change reliably after it wakes up when Broadcast() is called. In addition, for the two routines to communicate reliably, a mutex must be used, as several variables are changed by the two routines, e.g. numInFlightBlocks, lastCommittedBlock, etc. An atomic is not good for that as it does not have the "happen before" guaranties for multiple variables at once.

internal/replication/blockreplicator.go
tock-ibm: ... and I did forget to remove the TODO in line 345 ;-) which I'll fix

tock-ibm:
> Is there a way to test this limit behaviour?

Only by catching Log messages when the node reaches the Max, or by adding observable statistics that record the number of times the node gets there... I'll add a test that does the first approach.

gennadylaventman (approved):

gennadylaventman (approved):

PR #227, opened 2021-10-31
Augmented header

gennadylaventman:


internal/blockstore/commit_and_query.go
```diff
@@ -268,11 +269,21 @@ func (s *Store) storeBlockHeaders(header *types.BlockHeader) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "can't marshal block tx ids {%d, %v}", number, blockTx)
+	}
+
 	batch := &leveldb.Batch{}
 	batch.Put(constructHeaderBaseHashKey(number), blockHeaderBaseHash)
 	batch.Put(constructHeaderHashKey(number), blockHash)
 	batch.Put(constructHeaderBytesKey(number), blockHeaderBytes)
 	batch.Put(constructHeaderHashIndexKey(blockHash), encodeOrderPreservingVarUint64(number))
+	batch.Put(constructBlockTxKey(number), txIdBytes)
```
cendhu, line 286: Please rename all `BlockTx` to `BlockTxsID`. Initially, I got confused that we are storing the whole transaction. Let's be explicit.

protos/block_and_transaction.proto
```diff
@@ -227,3 +227,12 @@ message ConsensusMetadata {
...
   uint64 raft_index = 2;
 }
+
+message BlockTxIds {
+  repeated string tx_id = 1;
+}
+
+message AugmentedBlockHeader {
+  BlockHeader header = 1;
+  BlockTxIds tx_ids = 2;
```
cendhu, line 237: I would prefer to use the existing `BlockHeader` message type itself with the additional `repeated string tx_ids = 1`. The additional field would be empty when augmented is set to false. For marshalling, we can use some internal `BlockTxIds` proto at the `blockstore` package rather than using two `tx_ids`. In fact the `repeated string tx_id` should be `tx_ids`. I see it as a hack to avoid `GetTxIDs().GetTxIDs()`. Even the current usage of `GetTxIDs().GetTxID()[0]` is not good

pkg/types/block.go
```diff
@@ -0,0 +1,29 @@
...
+		txIDs = append(txIDs, m.GetDbAdministrationTxEnvelope().Payload.TxId)
+
+	case *Block_UserAdministrationTxEnvelope:
+		txIDs = append(txIDs, m.GetUserAdministrationTxEnvelope().Payload.TxId)
+
+	default:
+		return nil, errors.Errorf("unknown block payload")
+	}
+	return &BlockTxIds{TxId: txIDs}, nil
+}
```
cendhu, line 29: There is already a block to txIDs in util pkg. https://github.com/hyperledger-labs/orion-server/blob/main/internal/httputils/blocks.go

internal/httphandler/ledger_request_handler.go
```diff
@@ -33,6 +33,8 @@ func NewLedgerRequestHandler(db bcdb.DB, logger *logger.SugarLogger) http.Handle
 		logger:      logger,
 	}
 
+	// HTTP GET "/ledger/block/{blockId}?augmented=true" gets augmented block header
+	handler.router.HandleFunc(constants.GetBlockHeader, handler.blockQuery).Methods(http.MethodGet).Queries("augmented", "{isAugmented:true|false}")
 	// HTTP GET "/ledger/block/{blockId}" gets block header
 	handler.router.HandleFunc(constants.GetBlockHeader, handler.blockQuery).Methods(http.MethodGet)
```
cendhu, line 39: I think, we should use a single REST API endpoint. When setting `isAugmented=false`, we get the old behaviour.

pkg/types/block.go
gennadylaventman: I missed it, will change

internal/httphandler/ledger_request_handler.go
tock-ibm: or when when this key isAugmented is missing, resort to old behavior

gennadylaventman: This is exactly how it works. Handler route in line 37 just indicated that we want to look for query parameter `augmented`.  Leaving only route in line 39 ignores all query parameters...

internal/blockstore/commit_and_query_test.go
```diff
@@ -418,6 +423,40 @@ func TestTxValidationInfo(t *testing.T) {
 	})
 }
 
+func TestGetAugmentedHeader(t *testing.T) {
+
```
cendhu, line 427: Remove this line

internal/blockstore/location.proto
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 syntax = "proto3";
 
-option go_package = "github.ibm.com/blockchaindb/server/pkg/blockstore";
+option go_package = "github.com/hyperledger-labs/orion-server/internal/blockstore";
```
cendhu, line 5: Good catch

internal/httphandler/ledger_request_handler_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	"path"
 	"testing"
 
+	"github.com/golang/protobuf/proto"
+
```
cendhu, line 15: Remove this line

cendhu (approved):
Thanks @gennadylaventman

PR #240, opened 2021-11-04
Unit test flake

tock-ibm:
Eliminate flake by removing sleep, use zap.Hooks instead.


cendhu (approved):
I copied this approach to delivery service retry at the sdk.

PR #256, opened 2021-11-16
encoder for value

cendhu:
This commit adds an utility that can encode the given data to
base64 string. Note that this utility accepts either a JSON or
string value only.

### DIFF
--- a/cmd/base64_encoder/encoder.go
+++ b/cmd/base64_encoder/encoder.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"encoding/base64"
+	"flag"
+	"fmt"
+)
+
+var help = "The data field accepts a JSON or a string value. THE data flag must be set. Two example commands are shown below: \n\n" +
+	"  encoder -data='{\"userID\":\"admin\"}' \n" +
+	"  encoder -data='value' \n"
+
+func main() {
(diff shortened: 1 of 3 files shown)


cmd/base64_encoder/encoder.go
```diff
@@ -0,0 +1,26 @@
...
+	"flag"
+	"fmt"
+)
+
+var help = "The data field accepts a JSON or a string value. THE data flag must be set. Two example commands are shown below: \n\n" +
+	"  encoder -data='{\"userID\":\"admin\"}' \n" +
+	"  encoder -data='value' \n"
+
+func main() {
+	data := flag.String("data", "", "json or string data to be encoded. Surround the JSON data with single quotes. An example json data is '{\"userID\":\"admin\"}'")
```
tock-ibm, line 14: why use a flag if there is only one input?
do you plan or foresee the need to use additional flags in the future? maybe different encodings? i.e. base64std, vs base64url, etc?

tock-ibm (commented):
overall looks good, just one question about the structure of the input and the flags

cmd/base64_encoder/encoder.go
cendhu: This is done for the consistency with other utilities such as `signer`, `bdb`.

tock-ibm (approved):

PR #261, opened 2021-11-16
Storing all sample configuration files to same folder

gennadylaventman:
Configuration files for local executable, for docker and all certificates moved to same folder (deployment),
subfolder for each type.

Updated docker to have 3 separate volumes - one for configuration, one for crypto and one for ledger.


cendhu (commented):
Can you please do `git mv` on the `crypto` rather than creating a new one. Because, I do not want to change all the `cURL` example commands again.

cendhu (approved):

PR #265, opened 2021-11-18
decoder decodes data response envelope

cendhu:
This commits add a decoder utility that would decode
a given json data response envelope.

It will get updated to decode the bas64 encoded value
in the subsequent PR.

### DIFF
--- a/cmd/base64_decoder/decoder.go
+++ b/cmd/base64_decoder/decoder.go
@@ -1,0 +1,62 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+
+	"github.com/hyperledger-labs/orion-server/pkg/types"
+)
+
+var help = "The decoder decodes the base64 encoded value field in the " +
+	"GetDataResponseEnvelope. Pass the json output of " +
+	"GetDataResponseEnvelope to the `-getresponse` flag"
+
+type GetDataResponseEnvelope struct {
+	Response  *GetDataResponse `json:"response,omitempty"`
+	Signature []byte           `json:"signature,omitempty"`
+}
+
+type GetDataResponse struct {
+	Header   *types.ResponseHeader `json:"header,omitempty"`
+	Value    string                `json:"value,omitempty"`
+	Metadata *types.Metadata       `json:"metadata,omitempty"`
+}
+
+func main() {
+	getresponse := flag.String(
+		"getresponse",
+		"",
+		"json output of GetDataResponseEnvelope. The value field in the json output will be decoded. Surround the JSON data with single quotes.",
+	)
+
+	flag.CommandLine.Output()
+	flag.Parse()
+
+	if *getresponse == "" {
+		fmt.Println(help)
+		flag.PrintDefaults()
+		return
+	}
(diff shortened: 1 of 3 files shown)


cmd/base64_decoder/decoder_test.go
```diff
@@ -0,0 +1,41 @@
...
+			require.NoError(t, err)
+
+			expectedOut := help + "\n  -getresponse string\n    \tjson output of " +
+				"GetDataResponseEnvelope. The value field in the json output will be decoded. " +
+				"Surround the JSON data with single quotes.\n"
+			require.Equal(t, expectedOut, string(out))
+		}
+	})
+
+	t.Run("correct args", func(t *testing.T) {
```
tock-ibm, line 26: do a test with a pipe, i.e. 
echo 'data' | decoder.go ... etc

cmd/base64_decoder/decoder.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+var help = "The decoder decodes the base64 encoded value field in the " +
+	"GetDataResponseEnvelope. Pass the json output of " +
+	"GetDataResponseEnvelope to the `-getresponse` flag"
+
+type GetDataResponseEnvelope struct {
+	Response  *GetDataResponse `json:"response,omitempty"`
+	Signature []byte           `json:"signature,omitempty"`
+}
```
tock-ibm, line 18: If you refrain from printing the signature, you can get away from redefining this, no?

tock-ibm (commented):
A couple of minor comments

cmd/base64_decoder/decoder_test.go
cendhu: Pipe would not work. We need to read the output into a variable and then pass on the variable when decoder is used with the `cURL` as shown below:

 ```
curl \
     -H "Content-Type: application/json" \
     -H "UserID: bob" \
     -H "Signature: MEUCIQDm6dLmAdd0X49JygTiUkh+brZxprWSr2+hcAH+QIu3AAIgF+m7kO33YXyyqSbnXS9HR79wt/aL3JGhKvXFQaFBJms=" \
     -X GET http://127.0.0.1:6001/data/db2/key1 | read json; ./base64_decoder -getresponse=$json | jq .
```

cmd/base64_decoder/decoder.go
cendhu: I know it is imperfect but removing the envelope field requires more unnecessary descriptions to this utility and documentation. On the other hand, this signature is invalid on the decoded data. 

I would like to keep this simple. The decoder just takes the `Envelope` and decodes the `value` field, output the envelope again with the decoded value field.

cmd/base64_decoder/decoder_test.go
tock-ibm: ok, that works

cmd/base64_decoder/decoder.go
tock-ibm: ok

tock-ibm (approved):

PR #296, opened 2021-12-08
Cluster bootstrap join in replication

tock-ibm:
A new node in the cluster bootstraps with a join block, which is a config block that includes it in the Nodes and Members arrays.
    
The new nodes initializes comm.HTTPTransport with the peers from the join block, and therefore  can reach all the current active peers.
    
The node starts by on-boarding, that is, fetching all the blocks from 1 (genesis) and the  join-block number (say, J). As it receives blocks, it commits them to the ledger and databases,     rebuilding the state. If the node crashes befor reaching block J, it will resume on-boarding from where it stopped.
    
When the node reaches the join-block, it uses the cluster config within to construct a raftpb.ConfState,     it reads the raft Index and Term, and initializes the raft-storage (WAL & memory) to signal to raft     that it had caught up until said raft.Index, and saves a snapshot. 
    
It then starts the raft node and continues as normal.

### DIFF
--- a/internal/replication/blockreplicator.go
+++ b/internal/replication/blockreplicator.go
@@ -6,6 +6,8 @@ package replication
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -51,7 +53,7 @@ type BlockReplicator struct {
 	proposeCh       chan *types.Block
 	raftID          uint64
 	raftStorage     *RaftStorage
-	raftNode        raft.Node
+	raftConfig      *raft.Config
 	oneQueueBarrier *queue.OneQueueBarrier // Synchronizes the block-replication deliver with the block-processor commit
 	transport       *comm.HTTPTransport
 	ledgerReader    BlockLedgerReader
@@ -62,10 +64,12 @@ type BlockReplicator struct {
 	doneProposeCh chan struct{}
 	doneEventCh   chan struct{}
 
-	// shared state between the propose-loop go-routine and event-loop go-routine
+	// shared state between the propose-loop go-routine and event-loop go-routine; as well as transport go-routines.
 	mutex                           sync.Mutex
 	clusterConfig                   *types.ClusterConfig
 	joinExistingCluster             bool
+	finishedJoin                    bool
+	raftNode                        raft.Node
 	lastKnownLeader                 uint64
 	lastKnownLeaderHost             string // cache the leader's Node host:port for client request redirection
 	cancelProposeContext            func() // cancels the propose-context if leadership is lost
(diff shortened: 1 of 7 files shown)


internal/replication/env_test.go
```diff
@@ -2,13 +2,15 @@ package replication_test
 
 import (
 	"fmt"
+	"github.com/hyperledger-labs/orion-server/internal/httputils"
```
gennadylaventman, line 5: goimports

tock-ibm: done

internal/replication/blockreplicator_reconfig_test.go
```diff
@@ -313,16 +304,93 @@ func TestBlockReplicator_ReConfig_AddPeer(t *testing.T) {
...
+
+	t.Log("Closing")
+	for _, node := range env.nodes {
+		err := node.Close()
+		require.NoError(t, err)
+	}
+}
+
+// Scenario: add a peer to the cluster, with frequent snapshots
+// - start 3 nodes, configured to tak frequent snapshots, wait for leader,
```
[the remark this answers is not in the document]
tock-ibm: done

internal/replication/blockreplicator.go
```diff
@@ -246,6 +269,68 @@ func (br *BlockReplicator) Start() {
 	<-readyProposeCh
 }
 
+// startOnBoarding pulls the missing blocks from the current ledger height up to (and including) the join block.
+// It then constructs the raftpb.ConfState and raftpb.Snapshot from the cluster configuration of the join block and
+// stores it. This way, raft storage starts with the right state to join the cluster. In addition, if the node is
+// restarted, the snapshot indicates that the node needs to restart normally, rather than perform on-boarding again.
+func (br *BlockReplicator) startOnBoarding(readyCh chan<- struct{}) {
+	br.lg.Info("Starting the on-boarding process")
+	close(readyCh)
```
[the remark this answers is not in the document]
tock-ibm: You are right, this is useless, I removed it.

internal/replication/blockreplicator.go
```diff
@@ -246,6 +267,67 @@ func (br *BlockReplicator) Start() {
...
+		}
+	}
+
+	br.lg.Infof("On-boarding completed successfully, starting replication")
+	genesisBlock, err := br.ledgerReader.Get(1)
+	if err != nil {
+		br.lg.Panicf("Failed to read genesis block after on-boarding")
+	}
+	genesisConfig := genesisBlock.GetConfigTxEnvelope().GetPayload().GetNewConfig()
+	br.lg.Debugf("Genesis config: %v", genesisConfig)
```
gennadylaventman, line 294: Why do we need access genesis block config here, I don't see it used anywhere. If it for debug purpose, lets enclosure it in some `if`

tock-ibm: That is correct. It is part of debug code I forgot to remove.

gennadylaventman (approved):

PR #297, opened 2021-12-09
Admin API to get last config block

tock-ibm:
When an admin joins a new server to the cluster, he has to provide to the new server
the last config block of the cluster. In this commit we expose an API that allows him
to get this block from the cluster.


gennadylaventman (commented):
This PR return or last config block or config block by number. Isn't it reasonable to have API to get effective config block for specific block number?

tock-ibm:
> This PR return or last config block or config block by number. Isn't it reasonable to have API to get effective config block for specific block number?

Provenance can be used to fetch the previous (from last) or next (from genesis), etc.
So lets keep this API simpler for now.

gennadylaventman (approved):

PR #302, opened 2021-12-16
Separate tx validation from block processor

tock-ibm:
In preparation for implementing config-tx validation pre-order (future commit), we separate the tx-validation logic into a separate package.
    - All validation code was moved to a new package
    - Respective types & methods used in other packages made Public (e.g. ValidateBlock)
    - The block processor consumes a single method during commit: 'ValidateBlock'.
    - Generation of Validator extracted from inside the block processor, now in transaction processor
    - Validator now can be shared with other components (future commit)
    
The block processor consumes a single method during commit: 'ValidateBlock'.
    
This will make it easier to expose validation logic to pre-ordering components.
    
This commit presents **no functional change**, just refactoring of existing code.

### DIFF
--- a/internal/txvalidation/config_tx_validator.go
+++ b/internal/txvalidation/config_tx_validator.go
@@ -1,7 +1,7 @@
 // Copyright IBM Corp. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package blockprocessor
+package txvalidation
 
 import (
 	"fmt"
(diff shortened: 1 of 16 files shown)


internal/txvalidation/config_tx_validator.go
```diff
@@ -27,7 +27,7 @@ type configTxValidator struct {
 	logger          *logger.SugarLogger
 }
 
-func (v *configTxValidator) validate(txEnv *types.ConfigTxEnvelope) (*types.ValidationInfo, error) {
+func (v *configTxValidator) Validate(txEnv *types.ConfigTxEnvelope) (*types.ValidationInfo, error) {
```
gennadylaventman, line 30: Why only in `configTxValidator` method `Validate` defined as public?

tock-ibm: Right now we don't have immediate plans to do the validation of other type of TXs pre-order, only config-TX. The validation of data TXs pre-order for example, may occur in the tx-reorderer component for example. I think that when we get to it, we'll have to review the requirements again, and maybe require some implementation changes.  Therefore I did not  want to make these methods public at this point.

The `configTxValidator.Validate`, on the other hand, will be used in the next PR that I plan.

gennadylaventman (approved):

PR #303, opened 2021-12-19
Integration testing: cluster node recovery

tohar-ayash:
Signed-off-by: tohar-ayash <toharayash1@gmail.com>

### DIFF
--- a/pkg/constants/http.go
+++ b/pkg/constants/http.go
@@ -106,6 +106,12 @@ func URLForGetConfig() string {
 	return GetConfig
 }
 
+// URLForGetClusterStatus returns url for GET request to find
+// status of the cluster
+func URLForGetClusterStatus() string {
+	return GetClusterStatus
+}
+
 func URLForLedgerBlock(blockNum uint64, augmented bool) string {
 	if augmented {
 		return LedgerEndpoint + fmt.Sprintf("block/%d?augmented=%t", blockNum, augmented)
--- a/test/cluster/basic_test.go
+++ b/test/cluster/basic_test.go
@@ -5,8 +5,10 @@ package cluster
 
 import (
 	"fmt"
+	"github.com/hyperledger-labs/orion-server/pkg/types"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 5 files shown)


test/cluster/basic_test.go
```diff
@@ -101,3 +101,348 @@ func TestBasicCluster(t *testing.T) {
...
+			followerValue = i
+		} else {
+			t.Logf("tx submitted: %s, %+v", txID, rcpt)
+			txOK++
+			leaderServer = srv
+			t.Logf("Leader is: %s", srv.URL())
+		}
+	}
+	require.Equal(t, 1, txOK)
+	require.Equal(t, 2, txRedirected)
```
tock-ibm, line 235: This part repeats itself in many tests, and is basically "submit to detect a leader". It can be done across all servers, or possibly a subset when some servers are knows to be down. You can put it in a function.
If there is no leader at all; for instance it takes some time to elect, then the function will fail, but you can then retry to submit-to-detect-a-leader until it eventually succeeds... for example, retry every 1s for 30s total. This can be useful when you do the rolling restarts tests. 

You can also use `assert.Eventually(t, isLeaderCond, 30*time.Second, 1*time.Second)`  to automate this condition, see e.g. https://github.com/hyperledger-labs/orion-server/blob/a5db614f1044993886df4719e3ac14d9a65aa5ad/internal/replication/blockreplicator_3node_test.go#L64

test/cluster/basic_test.go
```diff
@@ -101,3 +101,348 @@ func TestBasicCluster(t *testing.T) {
...
+	c, err := setup.NewCluster(setupConfig)
+	require.NoError(t, err)
+	defer c.ShutdownAndCleanup()
+
+	require.NoError(t, c.Start())
+
+	time.Sleep(10 * time.Second) //TODO replace by a call to detect a leader
+
+	var txOK int
+	leaders := [3]int{0, 0, 0}
```
tock-ibm, line 316: you use `int` as a `bool` right? why not just `leaders := [3]bool{false, false, false}` ? seems more readable

test/cluster/basic_test.go
```diff
@@ -101,3 +101,348 @@ func TestBasicCluster(t *testing.T) {
...
+					break
+				}
+			}
+		}
+		require.NoError(t, c.ShutdownServer(c.Servers[currentLeader]))
+		if oldLeader != -1 {
+			require.NoError(t, c.StartServer(c.Servers[oldLeader]))
+		}
+
+		time.Sleep(10 * time.Second)
```
tock-ibm, line 345: maybe it takes longer to elect a leader, so you can use the "eventually" approach, see above

tohar-ayash: Done

test/cluster/basic_test.go
tohar-ayash: Done

test/cluster/basic_test.go
```diff
@@ -101,3 +101,348 @@ func TestBasicCluster(t *testing.T) {
...
+			return errors.Errorf("Redirect blocked in test client: url: '%s', referrer: '%s', #via: %d", req.URL, req.Referer(), len(via))
+		},
+	}
+	c, err := setup.NewCluster(setupConfig)
+	require.NoError(t, err)
+	defer c.ShutdownAndCleanup()
+
+	require.NoError(t, c.Start())
+
+	require.NoError(t, c.Restart())
```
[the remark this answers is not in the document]
tohar-ayash: Done

test/cluster/basic_test.go
```diff
@@ -101,3 +101,348 @@ func TestBasicCluster(t *testing.T) {
...
+
+	require.NoError(t, c.ShutdownServer(leaderServer))
+
+	txID, rcpt, err := followerServer.WriteDataTx(t, worldstate.DefaultDBName, "bob", []byte{uint8(followerValue), uint8(followerValue)})
+	if err != nil {
+		t.Logf("txID: %s, error: %s", txID, err)
+	} else {
+		t.Logf("tx submitted: %s, %+v", txID, rcpt)
+	}
+
```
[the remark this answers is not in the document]
tohar-ayash: Done

test/cluster/basic_test.go
```diff
@@ -101,3 +101,348 @@ func TestBasicCluster(t *testing.T) {
...
+	require.NoError(t, c.StartServer(followerServer))
+
+	time.Sleep(10 * time.Second)
+
+	dataEnv, err := followerServer.QueryData(t, worldstate.DefaultDBName, "bob")
+	require.NoError(t, err)
+	dataResp := dataEnv.GetResponse()
```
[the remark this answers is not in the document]
tohar-ayash: Done

pkg/constants/http.go
```diff
@@ -106,6 +106,12 @@ func URLForGetConfig() string {
 	return GetConfig
 }
 
+// URLForGetClusterStatus returns url for GET request to find
+// status of the cluster
+func URLForGetClusterStatus() string {
+	return GetClusterStatus
+}
```
tock-ibm, line 113: This does nothing bu return a public constant... just use the constant

test/cluster/basic_test.go
```diff
@@ -5,8 +5,10 @@ package cluster
 
 import (
 	"fmt"
+	"github.com/hyperledger-labs/orion-server/pkg/types"
```
tock-ibm, line 8: goimports

test/cluster/basic_test.go
```diff
@@ -101,3 +103,424 @@ func TestBasicCluster(t *testing.T) {
...
+	t.Logf("data: %+v", dataResp)
+}
+
+// Scenario:
+// - start 3 servers in a cluster.
+// - wait for one to be the leader.
+// - submit 2 txs
+// - shutting down and starting all servers in the cluster.
+// - make sure the servers is in sync with the txs made before they were shut down.
+func TestClusterRecovery(t *testing.T) {
```
tock-ibm, line 278: rename to ClusterRestart

test/cluster/basic_test.go
```diff
@@ -101,3 +103,424 @@ func TestBasicCluster(t *testing.T) {
...
+		},
+	}
+	c, err := setup.NewCluster(setupConfig)
+	require.NoError(t, err)
+	defer c.ShutdownAndCleanup()
+
+	require.NoError(t, c.Start())
+
+	//find the leader
+	var clusterStatusResEnv *types.GetClusterStatusResponseEnvelope
```
tock-ibm, line 301: The check that all servers agree on a leader should be part of the "Eventually", that is you poll untill they all give a consistent answer. See for example [this function ](https://github.com/hyperledger-labs/orion-server/blob/67a69d1a6e0bb46a53590e1538a37d7417f8d35b/internal/replication/env_test.go#L345) and how  it is used in the replication unit tests.

test/cluster/basic_test.go
```diff
@@ -101,3 +103,424 @@ func TestBasicCluster(t *testing.T) {
...
+		t.Logf("txID: %s, error: %s", txID, err)
+	}
+
+	require.NoError(t, c.Restart())
+
+	//make sure all 3 nodes are active
+	require.Eventually(t, func() bool {
+		clusterStatusResEnv, err = c.Servers[leaderVal].QueryClusterStatus(t)
+		return clusterStatusResEnv != nil && len(clusterStatusResEnv.GetResponse().GetActive()) == 3
+	}, 30*time.Second, 100*time.Millisecond)
```
tock-ibm, line 331: also need to check here that they have an "AgreedLeader"

test/cluster/basic_test.go
```diff
@@ -101,3 +103,424 @@ func TestBasicCluster(t *testing.T) {
...
+
+// Scenario:
+// - start 3 servers in a cluster.
+// - wait for one to be the leader.
+// - stop the leader.
+// - wait for new server to be the leader.
+// - the leader submit a tx
+// - repeat until each server was a leader at least once.
+// - make sure each node submitted a valid tx as a leader
+func TestAllNodesGetLeadership(t *testing.T) {
```
tock-ibm, line 354: Lets split this into a second  PR on top of the simpler one

test/cluster/basic_test.go
```diff
@@ -101,3 +103,424 @@ func TestBasicCluster(t *testing.T) {
...
+	require.NotEqual(t, leaderVal, -1)
+	require.NotNil(t, leaderServer)
+
+	follower1 := (leaderVal + 1) % 3
+	follower2 := (leaderVal + 2) % 3
+
+	require.NoError(t, c.ShutdownServer(c.Servers[leaderVal]))
+	require.NoError(t, c.ShutdownServer(c.Servers[follower1]))
+
+	time.Sleep(10 * time.Second)
```
tock-ibm, line 488: get status and wait for eventually - only one node is active, also check leader is ""

test/cluster/basic_test.go
```diff
@@ -101,3 +103,424 @@ func TestBasicCluster(t *testing.T) {
...
+	var newLeader int
+	for i, srv := range c.Servers {
+		if i != follower1 {
+			require.Eventually(t, func() bool {
+				clusterStatusResEnv, err = srv.QueryClusterStatus(t)
+				return clusterStatusResEnv != nil && clusterStatusResEnv.GetResponse().GetLeader() != "" && err == nil
+			}, 30*time.Second, 100*time.Millisecond)
+			_, newLeader = c.GetServerByID(clusterStatusResEnv.GetResponse().GetLeader())
+		}
+	}
```
tock-ibm, line 507: again, write a utility function to return the agreed upon leader on a set of nodes. normally you use all, but now only two.  See replication unit tests for [example](https://github.com/hyperledger-labs/orion-server/blob/67a69d1a6e0bb46a53590e1538a37d7417f8d35b/internal/replication/blockreplicator_reconfig_test.go#L97).

pkg/constants/http.go
tohar-ayash: Done

test/cluster/basic_test.go
tohar-ayash: Done

test/cluster/basic_test.go
tohar-ayash: Done

PR #305, opened 2021-12-22
HTTPTransport reports active peers

tock-ibm:
1. Utilize the rafthttp transport method that reports active peers.
2. Register the raft endpoints correctly to the ServeMux.
3. Expose a new method comm.HTTPTransport.ActivePeers, this will be used in the API that reports the cluster status (future commit).
4. Correct unit tests in comm and replication in light of the fix in (2).

After registering the raft endpoints correctly (2), the nodes may connect in an asymmetric way, that is:
- say the shared cluster config defines: (node1  localhost:1001) (node2  localhost:1002) (node3 localhost:1003)
- and: node1 binds to 1001, node2 binds to 1002
- but: node3 binds to 1004
- then: the connectivity is asymmetric - node3 can reach node1 & node2 but not the other way around

This may happen when reconfiguring an endpoint: node3 from 1003 -> 1004; and if
- node3 restarts on the new port before the config-tx, or
- it is kept alive during config-tx and restarted later.

The correct procedure for node endpoint update is to stop the node, reconfigure the remaining, and start it on the new port.

The tests were updated accordingly.

### DIFF
--- a/internal/comm/httptransport_test.go
+++ b/internal/comm/httptransport_test.go
@@ -5,11 +5,14 @@ package comm_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -21,6 +24,7 @@ import (
 	"github.com/hyperledger-labs/orion-server/pkg/server/testutils"
 	"github.com/hyperledger-labs/orion-server/pkg/types"
 	"github.com/stretchr/testify/require"
+	"github.com/xiang90/probing"
 	"go.etcd.io/etcd/raft/raftpb"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
@@ -401,9 +405,13 @@ func TestNewHTTPTransport_TLS_FilePathFailure(t *testing.T) {
 
 // Scenario: update the endpoints of a peer.
 // - generate a test configuration for 3 servers.
-// - server 3 starts on a new port, not reflected yet in shared config. servers 1,2 will not be able to reach it.
+// - server 3 starts on a new port, not reflected yet in shared config.
+// - servers 1,2 will not be able to reach it, but server 3 will be able to reach them.
 // - servers are updated with the shared config, now messages get through.
 func TestHTTPTransport_UpdatePeers(t *testing.T) {
+	wg := sync.WaitGroup{}
+	wg.Add(3)
+
 	lg, err := logger.New(
 		&logger.Config{
 			Level:         "info",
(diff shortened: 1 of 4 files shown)


internal/comm/httptransport_test.go
```diff
@@ -425,40 +441,45 @@ func TestHTTPTransport_UpdatePeers(t *testing.T) {
...
-	err = tr2.SendConsensus([]raftpb.Message{{To: 3}})
-	require.NoError(t, err)
-	require.Eventually(t,
-		func() bool {
-			return cl2.ReportUnreachableCallCount() == 2
-		},
-		10*time.Second, 10*time.Millisecond,
-	)
+	wg.Wait()
+	time.Sleep(1 * time.Second)
```
gennadylaventman, line 451: I don't like unconditional sleep. Is any way to make it eventually?

internal/comm/httptransport_test.go
```diff
@@ -712,6 +746,121 @@ func TestHTTPTransport_RemovePeers(t *testing.T) {
...
+
+	tr5, _, err := startTransportWithLedger(t, lg, localConfigs, sharedConfig, 4, 0)
+	require.NoError(t, err)
+
+	require.Eventually(t, func() bool { return numActiveCond(tr4, 4, false) }, 10*time.Second, 1000*time.Millisecond)
+	require.Eventually(t, func() bool { return numActiveCond(tr5, 4, false) }, 10*time.Second, 1000*time.Millisecond)
+
+	for _, tr := range []*comm.HTTPTransport{tr1, tr2, tr3, tr4, tr5} {
+		activePeers := tr.ActivePeers(10*time.Millisecond, true)
+		require.Equal(t, "node5", activePeers["node5"].NodeId)
```
gennadylaventman, line 842: why not use loop:
```
for _, peerId := range []string{"node1", "node2", "node3", "node4", "node5"} {
       require.Equal(t, peerId, activePeers[peerId].NodeId)
}

File: internal/replication/blockreplicator_reconfig_test.go [hunk 3]
```diff
@@ -91,19 +110,14 @@ func TestBlockReplicator_ReConfig_Endpoint(t *testing.T) {
 
 		// submission to a follower will cause an error
 		err = env.nodes[follower1].blockReplicator.Submit(b)
-		assert.EqualError(t, err, expectedNotLeaderErr)
-		// submission to a partitioned node will cause an error
-		err = env.nodes[follower2].blockReplicator.Submit(b)
-		assert.EqualError(t, err, "not a leader, leader is RaftID: 0, with HostPort: ")
+		require.EqualError(t, err, expectedNotLeaderErr)
```
gennadylaventman, line 113: `follower2` updated couple of lines above, but not used anymore

internal/comm/httptransport_test.go
tock-ibm: Done

internal/comm/httptransport_test.go
tock-ibm: Done

internal/replication/blockreplicator_reconfig_test.go
tock-ibm: Done

gennadylaventman (approved):

PR #307, opened 2021-12-26
Get cluster status API

tock-ibm:
Add an API that allows any user to detect the configured nodes from ClusterConfig.Nodes
and the status of the nodes: who is the leader, and which nodes are active.


internal/bcdb/transaction_processor_test.go
```diff
@@ -565,6 +569,25 @@ func TestTransactionProcessor(t *testing.T) {
...
+	t.Run("get cluster status", func(t *testing.T) {
+		cryptoDir, conf := testConfiguration(t)
+		require.NotEqual(t, "", cryptoDir)
+		defer os.RemoveAll(conf.LocalConfig.Server.Database.LedgerDirectory)
+		env := newTxProcessorTestEnv(t, cryptoDir, conf)
+		defer env.cleanup()
+
+		setupTxProcessor(t, env, worldstate.DefaultDBName)
+
+		require.Eventually(t, func() bool {
```
gennadylaventman, line 582: Isn't it possible that leader will be elected fast enough and this condition never become true?

gennadylaventman (commented):
Some minor comments

pkg/constants/http.go
```diff
@@ -45,12 +45,14 @@ const (
...
-	GetConfig          = "/config/tx"
-	GetNodesConfig     = "/config/node"
-	GetNodeConfig      = "/config/node/{nodeId}"
+	ConfigEndpoint    = "/config/"
+	PostConfigTx      = "/config/tx"
+	GetConfig         = "/config/tx"
+	GetNodeConfigPath = "/config/node"
+	GetNodeConfig     = "/config/node/{nodeId}"
 	GetLastConfigBlock = "/config/block/last"
+	GetClusterStatus  = "/config/cluster"
```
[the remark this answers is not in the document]
tock-ibm: done

internal/bcdb/db.go
```diff
@@ -388,6 +402,56 @@ func (d *db) GetConfigBlock(querierUserID string, blockNumber uint64) (*types.Ge
...
+		Nodes:   nodes,
+		Version: configResponse.GetMetadata().GetVersion(),
+	}
+
+	leader, active := d.txProcessor.ClusterStatus()
+
+	// The configResponse is from the world-state-db, whereas the (leader, active) pair is from the block-replicator.
+	// Since they may be out of sync (if the request hits right in the middle of a config change), we make sure the
+	// response is self-consistent.
+	for _, activeID := range active {
```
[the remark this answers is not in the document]
tock-ibm: Hmm... I prefer not to mess with it now for saving a single loop... so let's leave it.

internal/bcdb/transaction_processor_test.go
tock-ibm: This condition asserts that the current node is the leader; that is when IsLeader() returns nil, the current node is leader, so it is fine.

gennadylaventman: I see

gennadylaventman (approved):

PR #311, opened 2022-01-04
Validate config preorder

tock-ibm:
Validate a config tx before proposing in to Raft.

We validate a config TX preorder to prevent a membership change to be applied to Raft, and then
have the TX marked invalid during commit. Note that ConfigTxValidator.Validate reads the current
config from the store. However, together with the code to prevent in-flight config blocks before
proposing a new one (future commit), the  config from the store should match the config stored in
the BlockReplicator.


internal/replication/blockreplicator.go
```diff
@@ -780,14 +789,31 @@ Propose_Loop:
...
+				// See: https://github.com/hyperledger-labs/orion-server/issues/281
+				valInfo, errVal := br.configTxValidator.Validate(configTxEnv)
+				if errVal != nil {
+					br.lg.Errorf("Failed to validate configTxEnv, internal error: %s", errVal)
+					br.releasePendingTXs(blockToPropose, "Declined to propose block, internal error in Validate", errVal)
+					continue Propose_Loop
+				}
+				if valInfo.Flag != types.Flag_VALID {
+					br.releasePendingTXs(blockToPropose, "Declined to propose block, invalid config tx",
+						&ierrors.BadRequestError{ErrMsg: fmt.Sprintf("Invalid config tx, reason: %s", valInfo.ReasonIfInvalid)})
```
gennadylaventman, line 807: Isn't `continue Propose_Loop` should be here as well?

tock-ibm: Good catch!

internal/replication/blockreplicator.go
```diff
@@ -139,17 +147,18 @@ func NewBlockReplicator(conf *Config) (*BlockReplicator, error) {
 		raftID:               raftID,
 		raftStorage:          storage,
 		oneQueueBarrier:      conf.BlockOneQueueBarrier,
+		transport:            conf.Transport,
```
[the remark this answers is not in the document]
tock-ibm: Formatting changed when I refactored, I think the new order matches the order of the fields in the struct.

gennadylaventman (approved):

PR #312, opened 2022-01-05
Prevent proposals inflight config

tock-ibm:
We are preventing all proposals when a config is in flight, by using the same mechanism that prevents proposals when there are too many regular blocks in flight. This is not strictly necessary (what we strictly need is to prevent proposing a config when another config is in-flight), but is easier to implement. The implications on steady state performance are negligible, as config transactions are rare.

### DIFF
--- a/internal/replication/blockreplicator.go
+++ b/internal/replication/blockreplicator.go
@@ -84,6 +84,7 @@ type BlockReplicator struct {
 	lastProposedBlockHeaderBaseHash []byte
 	lastCommittedBlock              *types.Block
 	numInFlightBlocks               uint32 // number of in-flight blocks
+	inFlightConfigBlockNumber       uint64 // the block number of the in-flight config, if any; 0 if none
 	condTooManyInFlightBlocks       *sync.Cond
 
 	appliedIndex uint64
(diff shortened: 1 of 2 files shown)


internal/replication/blockreplicator.go
```diff
@@ -948,20 +952,43 @@ func (br *BlockReplicator) updateLastProposal(lastBlockProposed *types.Block) {
 			br.lg.Panicf("Failed to compute last block base hash: %s", err)
 		}
 		br.numInFlightBlocks++
+		if httputils.IsConfigBlock(lastBlockProposed) {
+			if br.inFlightConfigBlockNumber > 0 {
+				br.lg.Panicf("A config block was proposed, but there is still a config block in-flight! proposed: %d, in-flight: %d",
+					lastBlockProposed.GetHeader().GetBaseHeader().GetNumber(), br.inFlightConfigBlockNumber)
+			}
```
gennadylaventman, line 959: Can't we, instead exiting on panic, somehow make it wait here, maybe for while?

tock-ibm: We could wait, but for what? the design makes sure that this never happens. If a config block was committed, then we should have been waiting on this exact number; not bigger or smaller number...

gennadylaventman (approved):

PR #314, opened 2022-01-06
Rename httputils package to utils

tock-ibm:
Rename `httputils` package to `utils`
Rename file `utils.go` to `http.go`

No functional change.

### DIFF
--- a/internal/utils/blocks_test.go
+++ b/internal/utils/blocks_test.go
@@ -1,13 +1,13 @@
 // Copyright IBM Corp. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package httputils_test
+package utils_test
 
 import (
 	"fmt"
 	"testing"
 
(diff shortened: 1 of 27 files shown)


internal/utils/blocks_test.go
```diff
@@ -1,13 +1,13 @@
 // Copyright IBM Corp. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package httputils_test
+package utils_test
```
gennadylaventman, line 4: Why do we need new package here?

gennadylaventman: Not related to this pr exactly, but you change package name in any case

tock-ibm: I like black box testing to check exported methods. 
Anyway, I changed it.

gennadylaventman: ok

gennadylaventman (approved):

PR #351, opened 2022-02-14
Cluster tests 

tohar-ayash:


test/cluster/basic_test.go
```diff
@@ -290,3 +291,319 @@ func TestNodeRecoveryLeader(t *testing.T) {
...
+
+//Scenario:
+// - start 3 servers in a cluster.
+// - wait for one to be the leader.
+// - stop the leader and one follower.
+// - submit a tx to the remaining server.
+// - there is no majority to pick leader => tx fails.
+// - restart one server so now there will be a majority to pick a leader.
+// - wait for one to be the leader.
+// - submit a tx => tx accepts.
```
tock-ibm, line 303: tx accepted.

test/cluster/basic_test.go
```diff
@@ -290,3 +291,319 @@ func TestNodeRecoveryLeader(t *testing.T) {
...
+		return c.AgreedHeight(t, 4, 0, 1, 2)
+	}, 30*time.Second, 100*time.Millisecond)
+}
+
+// Scenario:
+// - start 3 servers in a cluster.
+// - wait for one to be the leader.
+// - submit 2 txs
+// - shutting down and starting all servers in the cluster.
+// - make sure the servers is in sync with the txs made before they were shut down.
```
tock-ibm, line 446: are in sync

test/cluster/basic_test.go
```diff
@@ -290,3 +291,319 @@ func TestNodeRecoveryLeader(t *testing.T) {
...
+	}, 30*time.Second, 100*time.Millisecond)
+
+	require.Eventually(t, func() bool {
+		clusterStatusResEnv, err := c.Servers[newLeader].QueryClusterStatus(t)
+		return err == nil && clusterStatusResEnv != nil && len(clusterStatusResEnv.GetResponse().GetActive()) == 2
+	}, 30*time.Second, 100*time.Millisecond)
+
+	require.NoError(t, c.ShutdownServer(c.Servers[follower1]))
+
+	//only one server is active now => no majority to pick leader
```
tock-ibm, line 372: Before shutting down the 2nd server: 
1. When we do this, go from 2 to 1, we have two options: shutdown the leader, or the follower. Lets make sure we do both options... lets start by shutting down the follower, as this exposed the bug, when the leader has some uncommitted items in its log before it loses majority. I.e. find which of `follower1` & `follower2` is the new leader, and shut the other one.
2. Add another test case that does the opposite.

test/cluster/basic_test.go
```diff
@@ -290,3 +291,319 @@ func TestNodeRecoveryLeader(t *testing.T) {
...
+	require.Eventually(t, func() bool {
+		_, _, err = c.Servers[follower2].WriteDataTx(t, worldstate.DefaultDBName, "bob", []byte("bob-data"))
+		return err != nil && err.Error() == "failed to submit transaction, server returned: status: 503 Service Unavailable, message: Cluster leader unavailable"
+	}, 60*time.Second, 100*time.Millisecond)
```
tock-ibm, line 376: This part is actually what exposed the bug, as it will submit a tx into the leader just before it loses majority, and thus have some uncommitted items in its log, so this is good.

test/cluster/basic_test.go
```diff
@@ -290,3 +291,319 @@ func TestNodeRecoveryLeader(t *testing.T) {
...
+
+	//restart one server => 2 servers are active
+	require.NoError(t, c.StartServer(c.Servers[leaderIndex]))
```
tock-ibm, line 379: I suggest renaming all the leader follower1 follower2 with a suffix that numbers the "round", as these change when we shut down and restart. e.g. leaderRound1 follower1Round1 etc. It will be easier to follow I think...

test/cluster/basic_test.go
tock-ibm: (or this can be done in the same test when all three recover,, with some additional keys `eliyahu` `fiona` `george` etc)

test/cluster/basic_test.go
```diff
@@ -290,3 +291,319 @@ func TestNodeRecoveryLeader(t *testing.T) {
...
+			dataEnv, err := c.Servers[newLeader].QueryData(t, worldstate.DefaultDBName, key)
+			if dataEnv != nil && dataEnv.GetResponse().GetValue() != nil {
+				dataVal := dataEnv.GetResponse().GetValue()
+				require.Equal(t, dataVal, []byte(key+"-data"))
+				t.Logf("data: %+v", dataVal)
+			}
+			return dataEnv != nil && dataEnv.GetResponse().GetValue() != nil && err == nil
+		}, 30*time.Second, 100*time.Millisecond)
+
+	}
```
tock-ibm, line 517: close space

test/cluster/basic_test.go
```diff
@@ -290,3 +291,319 @@ func TestNodeRecoveryLeader(t *testing.T) {
...
+			currentLeader = c.AgreedLeader(t, (oldLeader+1)%3, (oldLeader+2)%3)
+			return currentLeader >= 0
+		}, 30*time.Second, 100*time.Millisecond)
+	}
+
+	require.NoError(t, c.StartServer(c.Servers[oldLeader]))
+	require.Eventually(t, func() bool {
+		currentLeader = c.AgreedLeader(t, (oldLeader+1)%3, (oldLeader+2)%3)
+		return currentLeader >= 0
+	}, 30*time.Second, 100*time.Millisecond)
```
tock-ibm, line 585: shouldn't this be on all 3 now?, i.e. all 3 servers are alive? -> `AgreedLeader(t, 0, 1, 2)`

test/cluster/basic_test.go
```diff
@@ -290,3 +291,319 @@ func TestNodeRecoveryLeader(t *testing.T) {
...
+
+	//make sure all 3 nodes are active
+	require.Eventually(t, func() bool {
+		clusterStatusResEnv, err := c.Servers[currentLeader].QueryClusterStatus(t)
+		return err == nil && clusterStatusResEnv != nil && len(clusterStatusResEnv.GetResponse().GetActive()) == 3
+	}, 30*time.Second, 100*time.Millisecond)
+
+	require.Eventually(t, func() bool {
+		return c.AgreedHeight(t, 4, 0, 1, 2)
+	}, 30*time.Second, 100*time.Millisecond)
```
tock-ibm, line 595: Are you sure the height will always be 4? what if the loop runs more times because a node is elected twice before all get elected?
Maybe add a counter to the loop to count how many TXs where actually submitted?

test/cluster/basic_test.go
```diff
@@ -290,3 +291,319 @@ func TestNodeRecoveryLeader(t *testing.T) {
...
+	currentLeader := -1
+	require.Eventually(t, func() bool {
+		currentLeader = c.AgreedLeader(t, 0, 1, 2)
+		return currentLeader >= 0
+	}, 30*time.Second, 100*time.Millisecond)
+
+	oldLeader := -1
+	for !leaders[0] || !leaders[1] || !leaders[2] {
+		leaders[currentLeader] = true
+		txID, rcpt, err := c.Servers[currentLeader].WriteDataTx(t, worldstate.DefaultDBName, strconv.Itoa(currentLeader), []byte{uint8(currentLeader)})
```
tock-ibm, line 561: I am not sure `currentLeader` is a good candidate to the key, as a node might be elected twice, and therefore overwrite the same key. maybe a loop counter? see comment bellow as well.

test/cluster/basic_test.go
tohar-ayash: Done

test/cluster/basic_test.go
tohar-ayash: Done

test/cluster/basic_test.go
tohar-ayash: Done

test/cluster/basic_test.go
tohar-ayash: Done

test/cluster/basic_test.go
tohar-ayash: Done

test/cluster/basic_test.go
tohar-ayash: Done

tock-ibm (approved):

PR #354, opened 2022-02-28
test user addition and acl on database

cendhu:

### DIFF
--- a/test/setup/server_setup.go
+++ b/test/setup/server_setup.go
@@ -61,7 +61,6 @@ type Server struct {
 
 // NewServer creates a new blockchain database server
 func NewServer(id uint64, clusterBaseDir string, baseNodePort, basePeerPort uint32, checkRedirect func(req *http.Request, via []*http.Request) error, logger *logger.SugarLogger) (*Server, error) {
-
 	sNumber := strconv.FormatInt(int64(id+1), 10)
 	s := &Server{
 		serverNum:           id + 1,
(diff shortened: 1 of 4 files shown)


test/user/user_db_acl_test.go
```diff
@@ -0,0 +1,192 @@
...
+	setupConfig := &setup.Config{
+		NumberOfServers:     1,
+		TestDirAbsolutePath: dir,
+		BDBBinaryPath:       "../../bin/bdb",
+		CmdTimeout:          10 * time.Second,
+		BaseNodePort:        nPort,
+		BasePeerPort:        pPort,
+		CheckRedirectFunc: func(req *http.Request, via []*http.Request) error {
+			return errors.Errorf("Redirect blocked in test client: url: '%s', referrer: '%s', #via: %d", req.URL, req.Referer(), len(via))
+		},
```
tock-ibm, line 58: when using one server you can leave this as nil `CheckRedirectFunc: nil`

test/user/user_db_acl_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	c, err := setup.NewCluster(setupConfig)
+	require.NoError(t, err)
+	defer c.ShutdownAndCleanup()
+
+	require.NoError(t, c.Start())
+	time.Sleep(10 * time.Second)
```
tock-ibm, line 65: do an eventually on the GetClusterStatus until there is a leader, like here: 
https://github.com/hyperledger-labs/orion-server/blob/08e8788b4c2ca55702870cfa94afefcc5395b110/test/cluster/basic_test.go#L139

test/user/user_db_acl_test.go
```diff
@@ -0,0 +1,192 @@
...
+					},
+				},
+			},
+		},
+	})
+
+	client, err := s.NewRESTClient(nil)
+	require.NoError(t, err)
+
+	t.Run("read from db1 would succeed", func(t *testing.T) {
```
tock-ibm, line 90: those are missing:
- write to db3 should fail
- write to db2 should succeed

also, update user:
- db privilege
- certificate

test/setup/server_setup.go
```diff
@@ -181,12 +180,47 @@ func (s *Server) QueryData(t *testing.T, db, key string) (*types.GetDataResponse
 	return response, err
 }
 
-func (s *Server) WriteDataTx(t *testing.T, db, key string, value []byte) (string, *types.TxReceipt, error) {
-	client, err := s.NewRESTClient(s.clientCheckRedirect)
+func (s *Server) QueryUser(t *testing.T, userID string) (*types.GetUserResponseEnvelope, error) {
```
tock-ibm, line 183: How about also query user `alice` by a user other than `admin`? e.g. `alice` herself, or `bob`?
We can then check access control on the user db itself.

tock-ibm (commented):
I realize this may not be ready for review, but O provide some early feedback anyway.

test/user/user_db_acl_test.go
cendhu: Done

test/user/user_db_acl_test.go
cendhu: Done

test/user/user_db_acl_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+var baseNodePort uint32
+var basePeerPort uint32
+var portMutex sync.Mutex
+
+func init() {
+	baseNodePort = 6100
+	basePeerPort = 7100
```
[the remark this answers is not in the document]
cendhu: Changed. How can we ensure uniqueness in the long term?

test/user/user_db_acl_test.go
cendhu: Done except the certificate part as it will be tested in the subsequent PR.

test/setup/server_setup.go
cendhu: This can be done in the subsequent PR.

test/user/user_db_acl_test.go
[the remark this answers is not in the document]
tock-ibm: maybe we need a central table in the layer above where we allocate ranges according to packages.

tock-ibm (approved):

PR #356, opened 2022-03-07
test user credential updates and delete

cendhu:


test/setup/cluster_setup.go
```diff
@@ -386,45 +387,59 @@ func (c *Cluster) createUsersCryptoMaterials() error {
...
+	pemUserKeyFile, err := os.Create(userKeyPath)
+	if err != nil {
+		return err
+	}
+	if _, err = pemUserKeyFile.Write(pemUserKey); err != nil {
+		return err
+	}
+	if err = pemUserKeyFile.Close(); err != nil {
+		return err
+	}
```
tock-ibm, line 430: you can use `os.WriteFile` t do all this in shorter way.

test/setup/cluster_setup.go
```diff
@@ -386,45 +387,59 @@ func (c *Cluster) createUsersCryptoMaterials() error {
...
-		}
+	return nil
+}
 
-		userKeyPath := path.Join(c.testDirAbsPath, "users", user+".key")
-		pemUserKeyFile, err := os.Create(userKeyPath)
+func (c *Cluster) CreateOrUpdateUserCerts(user string, keyPair tls.Certificate) error {
+	var err error
+	if reflect.ValueOf(keyPair).IsZero() {
+		keyPair, err = tls.X509KeyPair(c.rootCAPemCert, c.caPrivKey)
```
tock-ibm, line 399: This pattern is confusing... in other packages it seems it is always "Update", whereas internally in the test package it is always create. Can we provide two methods? they may map to an internal "ugly" method, but passing in an empty key as flag for Update is a bit obscure for my taste.

test/setup/cluster_setup.go
cendhu: Done

test/setup/cluster_setup.go
cendhu: Done

test/setup/cluster_setup.go
```diff
@@ -380,51 +388,68 @@ func (c *Cluster) createUsersCryptoMaterials() error {
...
+	// if err != nil {
+	// 	return err
+	// }
+	// _, err = pemUserCertFile.Write(pemUserCert)
+	// if err != nil {
+	// 	return err
+	// }
+	// if err = pemUserCertFile.Close(); err != nil {
+	// 	return err
+	// }
```
tock-ibm, line 425: please remove

test/setup/cluster_setup.go
```diff
@@ -380,51 +388,68 @@ func (c *Cluster) createUsersCryptoMaterials() error {
...
+	// pemUserKeyFile, err := os.Create(userKeyPath)
+	// if err != nil {
+	// 	return err
+	// }
+	// if _, err = pemUserKeyFile.Write(pemUserKey); err != nil {
+	// 	return err
+	// }
+	// if err = pemUserKeyFile.Close(); err != nil {
+	// 	return err
+	// }
```
tock-ibm, line 440: please remove

tock-ibm:
all looks good except for a couple of commented out code that was left behind

test/setup/cluster_setup.go
cendhu: Done

test/setup/cluster_setup.go
cendhu: Done

tock-ibm (approved):

PR #363, opened 2022-03-24
testing user tx error cases

cendhu:

### DIFF
--- a/test/user/user_error_cases_test.go
+++ b/test/user/user_error_cases_test.go
@@ -1,0 +1,147 @@
+package user
+
+import (
+	"io/ioutil"
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/hyperledger-labs/orion-server/pkg/constants"
+	"github.com/hyperledger-labs/orion-server/pkg/server/testutils"
+	"github.com/hyperledger-labs/orion-server/pkg/types"
+	"github.com/hyperledger-labs/orion-server/test/setup"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUserTxErrorCases(t *testing.T) {
+	dir, err := ioutil.TempDir("", "int-test")
+	require.NoError(t, err)
+
+	nPort, pPort := getPorts(1)
+	setupConfig := &setup.Config{
+		NumberOfServers:     1,
+		TestDirAbsolutePath: dir,
+		BDBBinaryPath:       "../../bin/bdb",
+		CmdTimeout:          10 * time.Second,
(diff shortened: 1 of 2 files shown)


test/user/user_error_cases_test.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		_, err := s.SubmitTransaction(t, constants.PostUserTx, &types.UserAdministrationTxEnvelope{
+			Payload:   userTx,
+			Signature: testutils.SignatureFromTx(t, s.AdminSigner(), userTx),
+		})
+		require.EqualError(t, err, "TxValidation: Flag: INVALID_DATABASE_DOES_NOT_EXIST, Reason: the database [db1] present in the db permission list does not exist in the cluster")
+	})
+
+	t.Run("creating an user with incorrect certificate would fail", func(t *testing.T) {
```
tock-ibm, line 67: Lets also try a valid certificate but from the wrong CA?

tock-ibm:
@cendhu A question not directly related to this PR: what happens when a user is onboarded correctly, but then, the CA that issued its certificate is removed from Orion? I think that the user will continue to be able to access Orion; Is that what we want?

test/user/user_error_cases_test.go
cendhu: I have created an issue to handle that separately https://github.com/hyperledger-labs/orion-server/issues/396

cendhu:
> @cendhu A question not directly related to this PR: what happens when a user is onboarded correctly, but then, the CA that issued its certificate is removed from Orion? I think that the user will continue to be able to access Orion; Is that what we want?

Only during the enrollment, Certificate is validated against the CA. For signature verification, we only fetch the user's certificate and do not check against the current CA. 

We can fix this by link each user certificate with the CA certificate. When the root CA changes, we can invalidate respective user's certificate (implicitly). Whenever we do a signature verification, we need to check whether the CA's version associated with this certificate has been changed. If so, throw an error. What do you think?

tock-ibm:
> 
We could probably do this when the CA is changed; otherwise it will be to costly, in the hot path of every request. That is, when we change the CA config (note there could be more than one root/intermediary), revalidate all users and put a flag on each one that is not valid. In addition, when we change CAs, we should make sure we don't implicitly invalidate the current admins, or we will lock the admin out of the DB. There is an issue for all that here: https://github.com/hyperledger-labs/orion-server/issues/154

tock-ibm:
@cendhu LGTM. please rebase.

tock-ibm (approved):

PR #389, opened 2022-04-28
Start server while on-boarding

tock-ibm:

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"time"
 
 	"github.com/hyperledger-labs/orion-server/config"
 	"github.com/hyperledger-labs/orion-server/internal/bcdb"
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -124,12 +125,18 @@ func New(conf *config.Configurations) (*BCDBHTTPServer, error) {
...
-	} else {
-		s.logger.Infof("Server starting at ledger height [%d]", blockHeight)
+	for {
+		if blockHeight, err := s.db.LedgerHeight(); err != nil {
+			return err
+		} else if blockHeight == 0 {
+			// If the server is bootstrapping from the peers, it might still be trying to get the genesis block; We
+			// wait for that to happen before we start serving requests.
+			s.logger.Infof("Ledger height =0, server boostraping from peers; waiting for height >0; going to sleep for 1s...")
+			time.Sleep(time.Second)
```
cendhu, line 135: Weak guarantee. We need to let the bootstrapping process know whether the peer is being added to an existing network such that it can retry for several times before failing. 

Please create a issue to handle this cleanly. For now, I am merging this PR to fix the bug.

cendhu (approved):

cendhu:
Need to be rebased

pkg/server/server.go
tock-ibm: The block replicator is trying and retrying forever in a different go-routine. This main thread is just holding off on starting to serve client requests until we have at least the genesis block. It is not clean... that is true.

tock-ibm: https://github.com/hyperledger-labs/orion-server/issues/397

cendhu (approved):

PR #394, opened 2022-05-02
provenance queries test: historical data api

cendhu:
This commit covers all historical data apis
1. GetAllValues()
2. GetValueAt()
3. GetNextValues()
4. GetPreviousValues()
5. GetMostRecentValueAtOrBelow()

### DIFF
--- a/internal/provenance/commit_and_query.go
+++ b/internal/provenance/commit_and_query.go
@@ -9,11 +9,11 @@ import (
 	"sort"
 	"strings"
 
-	interrors "github.com/hyperledger-labs/orion-server/internal/errors"
-	"github.com/hyperledger-labs/orion-server/pkg/types"
 	"github.com/cayleygraph/cayley"
 	"github.com/cayleygraph/cayley/graph"
 	"github.com/cayleygraph/quad"
+	interrors "github.com/hyperledger-labs/orion-server/internal/errors"
+	"github.com/hyperledger-labs/orion-server/pkg/types"
 	"github.com/pkg/errors"
 )
 
@@ -153,7 +153,7 @@ func (s *Store) addWrites(tx *TxDataForProvenance, batch graph.BatchWriter) erro
 			// old version would not have been passed if it was deleted in the worldstate database already
 			// but we can find the old version from the provenance store even if it was deleted already
 			s.logger.Debug("fetching last deleted version of key [" + actualKey + "] from db [" + tx.DBName + "]")
-			lastVer, err := s.getLastDeletedVersion(tx.DBName, write.Key)
+			lastVer, err := s.getLastDeletedVersion(tx.DBName, actualKey)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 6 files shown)


test/provenance/provenance_query_test.go
```diff
@@ -0,0 +1,506 @@
...
+				Value: []byte("key2value3"),
+				Metadata: &types.Metadata{
+					Version: &types.Version{
+						BlockNum: 7,
+						TxNum:    0,
+					},
+				},
+			},
+		}
+		resp, err = s.GetAllValues(t, "db1", "key2", "alice")
```
tock-ibm, line 159: What happens when bob tries to get all values? those are not his keys right?
This is a hole in our privacy... 
We should think about what to do with this before we go to full pledged access control.

Only admins? 
Ability to disable provenance?
Or wait for provenance ACLs...

test/provenance/provenance_query_test.go
```diff
@@ -0,0 +1,506 @@
...
+		require.ElementsMatch(t, expectedKey3Values, resp.GetResponse().GetValues())
+	})
+}
+
+// 1. Store key1 with open access to all users in database 1
+// 2. Store key2 with read-write access only for alice in database 1
+// 3. Store key3 with open acess to all users in database 2
+// 4. Bob reads key1 and update the same in database 1
+// 5. Alice reads key1 and key2, update key2 while deleting key1 in database 1
+// 6. Alice reads key2 from database 1 and key3 from database 2 and updates key2 and key3
```
tock-ibm, line 314: I am not sure this is enough data to cover all the use cases... maybe use a few of additional keys to create gaps between the versions?
Also, what happens when a key is written several time, and then deleted, and then added again?

internal/httphandler/provenance_request_handler_test.go
```diff
@@ -788,6 +789,7 @@ func assertTestCase(t *testing.T, tt testCase, responseType interface{}) {
 }
 
 func constructRequestForTestCase(t *testing.T, url string, query interface{}, signer crypto.Signer, signerID string) *http.Request {
+	fmt.Println(url)
```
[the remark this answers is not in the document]
cendhu: Done

test/provenance/provenance_query_test.go
```diff
@@ -0,0 +1,506 @@
+// Copyright IBM Corp. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package provenancetest
```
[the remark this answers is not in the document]
cendhu: Done

test/provenance/provenance_query_test.go
```diff
@@ -0,0 +1,506 @@
...
+	require.NoError(t, err)
+
+	_, err = s.SubmitTransaction(t, constants.PostDataTx, &types.DataTxEnvelope{
+		Payload:    dataTx,
+		Signatures: map[string][]byte{"alice": testutils.SignatureFromTx(t, aliceSigner, dataTx)},
+	})
+	require.NoError(t, err)
+
+	// 4. Bob reads key1 and update the same in database 1
+	resp, err := s.QueryData(t, "db1", "key1", "alice")
```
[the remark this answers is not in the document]
cendhu: Done

test/provenance/provenance_query_test.go
cendhu: Added scenario so that when we add ACL, a test will fail.

test/provenance/provenance_query_test.go
cendhu: Good question. Though we have a similar scenario in the unit-test, there is a bug in the production code. Fixed it in #399

tock-ibm (approved):

PR #399, opened 2022-05-09
fix backward link to deleted key in provenance

cendhu:
When the key is deleted from the world state in
a one block and the same key is added in the future
block, there should be two links (previous/next) between old deleted
value and the new value. Due to a mistake in the unit-test,
we missed to detect the bug in the production code.

Now with the integration test, we could detect this bug.
To get the last deleted value to construct the next/previous
link with the new value, we need to pass the correct key. The
existing code passes the incorrect key to find the last deleted
key node.

### DIFF
--- a/internal/provenance/commit_and_query.go
+++ b/internal/provenance/commit_and_query.go
@@ -9,11 +9,11 @@ import (
 	"sort"
 	"strings"
 
-	interrors "github.com/hyperledger-labs/orion-server/internal/errors"
-	"github.com/hyperledger-labs/orion-server/pkg/types"
 	"github.com/cayleygraph/cayley"
 	"github.com/cayleygraph/cayley/graph"
 	"github.com/cayleygraph/quad"
+	interrors "github.com/hyperledger-labs/orion-server/internal/errors"
+	"github.com/hyperledger-labs/orion-server/pkg/types"
 	"github.com/pkg/errors"
 )
 
@@ -153,7 +153,7 @@ func (s *Store) addWrites(tx *TxDataForProvenance, batch graph.BatchWriter) erro
 			// old version would not have been passed if it was deleted in the worldstate database already
 			// but we can find the old version from the provenance store even if it was deleted already
 			s.logger.Debug("fetching last deleted version of key [" + actualKey + "] from db [" + tx.DBName + "]")
-			lastVer, err := s.getLastDeletedVersion(tx.DBName, write.Key)
+			lastVer, err := s.getLastDeletedVersion(tx.DBName, actualKey)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


internal/provenance/commit_and_query_test.go
```diff
@@ -847,109 +837,109 @@ func TestGetNextValues(t *testing.T) {
...
-			dbName: "db1",
-			key:    "key3",
-			version: &types.Version{
-				BlockNum: 3,
-				TxNum:    0,
-			},
-			limit:          -1,
-			expectedValues: nil,
-		},
+		// {
```
tock-ibm, line 840: all the tests are commented out...

internal/provenance/commit_and_query.go
```diff
@@ -308,6 +308,7 @@ func (s *Store) getDeletedValuesWithoutLock(dbName, key string) ([]*types.ValueW
 	s.logger.Debugf("fetch all historical deleted values associated with the key [%s] in db [%s]", key, dbName)
 	cKey := constructCompositeKey(dbName, key)
 	p := cayley.StartPath(s.cayleyGraph, quad.String(cKey)).Out().Tag("deleted_value").In(quad.String(DELETES)).Back("deleted_value")
+	// p := cayley.StartPath(s.cayleyGraph, quad.String(cKey)).Out().In(quad.String(DELETES))
```
tock-ibm, line 311: what is this?

internal/provenance/commit_and_query_test.go
cendhu: Enabled them.

internal/provenance/commit_and_query.go
cendhu: Forgot to remove

tock-ibm:
```
block5TxsData := []*TxDataForProvenance{
		{
			IsValid: true,
			DBName:  "db1",
			UserID:  "user2",
			TxID:    "tx6",
			Writes: []*types.KVWithMetadata{
				{
					Key:   "key1",
					Value: []byte("value5"),
					Metadata: &types.Metadata{
						Version: &types.Version{
							**BlockNum: 4,**            //<<< should be 5
							TxNum:    0,
						},
					},
				},
			},
              }
```
and
```
	block6TxsData := []*TxDataForProvenance{
		{
			IsValid: true,
			DBName:  "db1",
			UserID:  "user2",
			TxID:    "tx50",
			Deletes: map[string]*types.Version{
				"key1": {
					**BlockNum: 4,**        //<<< should be 5 as well
					TxNum:    0,
				},
			},
		},
	}
```

cendhu:
Done

tock-ibm (approved):

PR #403, opened 2022-05-10
Ledger test: path

tock-ibm:


test/queries/ledger_test.go
```diff
@@ -186,3 +192,187 @@ func TestLedgerBlockQueries(t *testing.T) {
...
+			name:            "1 to 10",
+			start:           1,
+			end:             10,
+			expectZeroBased: []uint64{9, 8, 0},
+		},
+		{
+			name:            "5 to 15",
+			start:           5,
+			end:             15,
+			expectZeroBased: []uint64{14, 12, 8, 4},
```
cendhu, line 260: Using different expected values creates confusion. In the unit-test, we pass non-zero based value, i.e., 15, 13, 8, 5 but here we pass zero based and then add 1 in the assertion. Initially, I was wondering why block list is wrong and realised after looking at the `+1`. You can have the zero based value in the comment and actual value in the test.

test/queries/ledger_test.go
```diff
@@ -186,3 +192,187 @@ func TestLedgerBlockQueries(t *testing.T) {
...
+			name:            "1 to 1",
+			start:           1,
+			end:             1,
+			expectZeroBased: []uint64{0},
+		},
+		{
+			name:            "6 to 6",
+			start:           6,
+			end:             6,
+			expectZeroBased: []uint64{5},
```
cendhu, line 278: having 6 instead of 5 would make readability easier

tock-ibm: done

test/queries/ledger_test.go
tock-ibm: done

PR #412, opened 2022-05-19
integration test: Json queries

cendhu:


test/queries/json_queries_test.go
```diff
@@ -0,0 +1,453 @@
...
+	"time"
+
+	"github.com/hyperledger-labs/orion-server/pkg/constants"
+	"github.com/hyperledger-labs/orion-server/pkg/server/testutils"
+	"github.com/hyperledger-labs/orion-server/pkg/types"
+	"github.com/hyperledger-labs/orion-server/test/setup"
+	"github.com/stretchr/testify/require"
+)
+
+func TestJSONQueries(t *testing.T) {
```
tock-ibm, line 18: Test cases here look good. I am missing some negative/corner cases though.
- bad selector
- JSON query on a db w/o index
- an empty result set
- JSON query on db the user has no read permission
etc.

cendhu: Done

tock-ibm (approved):

PR #419, opened 2022-06-08
Tx proof test

tohar-ayash:


tock-ibm (approved):
This test only checks a single tx in a block. After introducing the facilities to asynch data TXs, that will result in multiple TXs in a block, this test must also be augmented with a test case where the block contains more than a single Tx.

PR #420, opened 2022-06-08
range queries with pagination

cendhu:
This commit adds range query on a given database.
`GET database?startkey=k1&endkey=k2&limit=10` would be used
by the client to range over a set of keys. If the
query record limit is reached, the response would
include the next start key.


internal/httphandler/data_request_handler.go
```diff
@@ -90,6 +96,43 @@ func (d *dataRequestHandler) dataQuery(response http.ResponseWriter, request *ht
...
+	data, err := d.db.GetDataRange(query.DbName, query.UserId, query.StartKey, query.EndKey)
+	if err != nil {
+		var status int
+
+		switch err.(type) {
+		case *errors.PermissionErr:
+			status = http.StatusForbidden
+		default:
+			status = http.StatusInternalServerError
+		}
```
tock-ibm, line 122: what about the internal BadRequest? 
e.g. when start > end?
i.e. what  would `q.db.GetIterator(dbName, startKey, endKey)` return in that case?

cendhu: It is a valid request and the response would be nil (as followed in many other databases)

protos/query.proto
```diff
@@ -39,6 +39,13 @@ message GetDataQuery {
   string key = 3;
 }
 
+message GetDataRangeQuery {
+  string user_id = 1;
+  string db_name = 2;
+  string start_key = 3;
+  string end_key = 4;
+}
+
```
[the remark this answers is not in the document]
cendhu: We don't need an envelope. It is not a post request but get with arguments as start, end, limit, etc...

config/config.go
```diff
@@ -78,6 +78,8 @@ type ServerConf struct {
 	Database DatabaseConf
 	// The lengths of various queues that buffer between internal components.
 	QueueLength QueueLengthConf
+	// QueryResponseLimit denotes the maximum number of records to be returned per range query
+	QueryResponseLimit int
```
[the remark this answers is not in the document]
cendhu: I replaced count based limit with size. Limit can be given by the user in the query.

internal/bcdb/db.go
```diff
@@ -255,11 +257,12 @@ func NewDB(conf *config.Configurations, logger *logger.SugarLogger) (DB, error)
...
 	worldstateQueryProcessor := newWorldstateQueryProcessor(
 		&worldstateQueryProcessorConfig{
-			nodeID:          localConf.Server.Identity.ID,
-			db:              levelDB,
-			blockStore:      blockStore,
-			identityQuerier: querier,
-			logger:          logger,
+			nodeID:             localConf.Server.Identity.ID,
+			db:                 levelDB,
+			queryResponseLimit: conf.LocalConfig.Server.QueryResponseLimit,
```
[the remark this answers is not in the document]
cendhu: Done

internal/bcdb/worldstate_query_processor.go
```diff
@@ -20,28 +20,31 @@ import (
...
 
 type worldstateQueryProcessor struct {
-	nodeID          string
-	db              worldstate.DB
-	blockStore      *blockstore.Store
-	identityQuerier *identity.Querier
-	logger          *logger.SugarLogger
+	nodeID             string
+	db                 worldstate.DB
+	queryResponseLimit int
```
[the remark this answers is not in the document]
cendhu: Done

internal/httphandler/utils.go
```diff
@@ -36,6 +36,13 @@ func extractVerifiedQueryPayload(w http.ResponseWriter, r *http.Request, queryTy
 			DbName: params["dbname"],
 			Key:    params["key"],
 		}
+	case constants.GetDataRange:
+		payload = &types.GetDataRangeQuery{
+			UserId:   querierUserID,
+			DbName:   params["dbname"],
+			StartKey: params["startkey"][1 : len(params["startkey"])-1],
+			EndKey:   params["endkey"][1 : len(params["endkey"])-1],
+		}
```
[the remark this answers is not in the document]
cendhu: There would not be a server side count limit. I have replaced it with size limit. The client can set the count limit to 0 for full scan or specify a limit to fetch only those number of records. At the SDK, if the limit is set to 0 by the client, the SDK would run pagination internally to fetch all entries till the user closes the iterator. When the limit is set, the SDK would still do pagination but only till the count limit specified by the user (no more results would be sent beyond this limit). In the later case, it is possible for the sdk to fetch a bit more records than needed.

internal/httphandler/utils.go
```diff
@@ -36,6 +36,29 @@ func extractVerifiedQueryPayload(w http.ResponseWriter, r *http.Request, queryTy
...
+			return nil, true
+		}
+
+		payload = &types.GetDataRangeQuery{
+			UserId:   querierUserID,
+			DbName:   params["dbname"],
+			StartKey: params["startkey"][1 : len(params["startkey"])-1],
+			EndKey:   params["endkey"][1 : len(params["endkey"])-1],
+			Limit:    uint32(limit),
+		}
```
tock-ibm, line 61: This assumes that there is always a "limit" "startkey" and "endkey". This could be made more flexible, i.e. no limit means 0; no endkey means endkey="", etc. This would also require changes to the (gorilla) matcher...

What do you think?

protos/query.proto
```diff
@@ -39,6 +39,14 @@ message GetDataQuery {
   string key = 3;
 }
 
+message GetDataRangeQuery {
+  string user_id = 1;
+  string db_name = 2;
+  string start_key = 3;
+  string end_key = 4;
+  uint32 limit = 5;
```
tock-ibm, line 47: uint64?

internal/httphandler/data_request_handler_test.go
```diff
@@ -240,6 +239,265 @@ func TestDataRequestHandler_DataQuery(t *testing.T) {
...
+	testCases := []struct {
+		name               string
+		requestFactory     func() (*http.Request, error)
+		dbMockFactory      func(response *types.GetDataRangeResponseEnvelope) bcdb.DB
+		expectedResponse   *types.GetDataRangeResponseEnvelope
+		expectedStatusCode int
+		expectedErr        string
+	}{
+		{
+			name: "valid get data range with a limit",
```
tock-ibm, line 275: What is the symbol for the "just before first"? is it an empty string?
What is the symbol for the "just after last"? is it an empty string?

I am missing an open range queries like 
start="" / end="key10"
start="key5" / end=""
etc.

internal/httphandler/data_request_handler.go
```diff
@@ -42,6 +42,13 @@ func NewDataRequestHandler(db bcdb.DB, logger *logger.SugarLogger) http.Handler
 		logger: logger,
 	}
 
+	rangeKeys := []string{
+		"startkey", "{startkey}",
+		"endkey", "{endkey}",
+		"limit", "{limit}",
+	}
```
tock-ibm, line 49: This assumes that all 3 parameters must exist, otherwise there is a 404, which is difficult to understand. This might be improved to be more flexible, by adding matchers to queries with default parameters and setting defaults. 
Or, add a "missing parameter" matcher that would give an informative response... e.g. `missing startkey, endkey, or limit` 

It is a bit weird to do `xxx?startkey=&endkey=&limit=0` in order to get the entire range no?
Or maybe it doesn't matter because it is all produced by the SDK?

internal/httphandler/utils.go
```diff
@@ -36,6 +36,29 @@ func extractVerifiedQueryPayload(w http.ResponseWriter, r *http.Request, queryTy
...
+				&types.HttpResponseErr{ErrMsg: "limit value " + strconv.Itoa(limit) + " should be greater than or equal to zero"},
+			)
+			return nil, true
+		}
+
+		payload = &types.GetDataRangeQuery{
+			UserId:   querierUserID,
+			DbName:   params["dbname"],
+			StartKey: params["startkey"][1 : len(params["startkey"])-1],
+			EndKey:   params["endkey"][1 : len(params["endkey"])-1],
```
tock-ibm, line 59: I am not sure what these two lines do... why chop the first char in the key?

protos/query.proto
cendhu: Done

internal/httphandler/utils.go
cendhu: Yes, in the future PRs. https://github.com/hyperledger-labs/orion-server/issues/423

internal/httphandler/utils.go
cendhu: the start and end key string would be double quoted as ?`startkey=""&endkey="key2"&limit=10`. Here, we are removing the quotes. Here, I follow the CouchDB range query syntax.

internal/httphandler/data_request_handler_test.go
cendhu: Done

internal/httphandler/data_request_handler.go
cendhu: > Or maybe it doesn't matter because it is all produced by the SDK?

Yes. However, we do support cURL based usage too. Hence, we can add that in the subsequent PRs.

internal/bcdb/worldstate_query_processor.go
```diff
@@ -126,6 +129,87 @@ func (q *worldstateQueryProcessor) getData(dbName, querierUserID, key string) (*
...
+		size += uint32(len(k) + len(itr.Value()))
+		if size > q.responseSizeLimit {
+			pendingResult = true
+			nextStartKey = k
+			if len(kvs) != 0 {
+				break
+			}
+
+			return nil, &errors.ServerRestrictionError{
+				ErrMsg: fmt.Sprintf("response size limit for queries is configured as %d bytes but a single record size itself is %d bytes. Increase the query response size limit at the server", q.responseSizeLimit, size),
```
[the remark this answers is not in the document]
cendhu: I agree. We need to have restriction on the key length and value length. https://github.com/hyperledger-labs/orion-server/issues/424

internal/bcdb/worldstate_query_processor.go
```diff
@@ -20,28 +20,31 @@ import (
...
 
 type worldstateQueryProcessor struct {
-	nodeID          string
-	db              worldstate.DB
-	blockStore      *blockstore.Store
-	identityQuerier *identity.Querier
-	logger          *logger.SugarLogger
+	nodeID            string
+	db                worldstate.DB
+	responseSizeLimit uint32
```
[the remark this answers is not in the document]
cendhu: Done

internal/bcdb/db.go
```diff
@@ -541,6 +545,25 @@ func (d *db) GetData(dbName, querierUserID, key string) (*types.GetDataResponseE
 	}, nil
 }
 
+// GetDataRange returns a range of values starting from the start key and till before the end key
+func (d *db) GetDataRange(dbName, querierUserID, startKey, endKey string, limit uint32) (*types.GetDataRangeResponseEnvelope, error) {
```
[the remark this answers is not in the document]
cendhu: Done

config/config.go
```diff
@@ -118,6 +120,11 @@ type QueueLengthConf struct {
 	Block                     uint32
 }
 
+// QueueProcessingConf holds the configuration associated with rich and range query processing
+type QueryProcessingConf struct {
+	ResponseSizeLimit uint32
+}
```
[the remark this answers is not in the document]
cendhu: Done

tock-ibm (approved):

Issue #428, opened 2022-06-28
Flake on integration test / replication unit test


tock-ibm:
**Hypothesis:**
Replication unit test fails silently, doesn't close, leaves a port bound.
Integration test fails to bind that port.

PR #430, opened 2022-06-29
Destroys cluster environment when a test fails

liran-funaro:
Solves #428 

This commit adds a `destroyClusterEnv()` method that closes all the nodes that were started and removes the environment directory.
Each test, calls `defer destroyClusterEnv(t, env)` right after the environment is created to make sure it will be destroyed
even in the case of test failure.

### DIFF
--- a/test/setup/server_setup.go
+++ b/test/setup/server_setup.go
@@ -17,7 +17,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


test/setup/server_setup.go
```diff
@@ -1117,12 +1118,25 @@ func (s *Server) shutdown() error {
...
+		if err == nil {
+			break
+		}
+		if _, ok := err.(*exec.ExitError); !ok {
+			return errors.Wrapf(err, "error while shutting down %s", s.serverID)
+		}
+	case <-time.After(10 * time.Second):
+		if err := s.cmd.Process.Kill(); err != nil {
+			return errors.Wrapf(err, "error while forcing shutting down %s", s.serverID)
+		}
```
tock-ibm, line 1139: This is unrelated to the destroy issue, and belongs to another package. lets do this in another PR?
I think it is ok to Kill, because we do not yet support graceful shutdown. IE we do not catch the signals and do different things for different signals. I am also not sure it is causing any problems to just Wait().

tock-ibm (approved):

PR #436, opened 2022-07-17
Provenance disable switch

tock-ibm:
Add a config option in the local config to disable the provenance store.
    
- When disabled:
  - The provenance store is not initialized and is set to nil.
  - In the commit path, no data is committed to the provenance store.
  - Provenance queries return 503 Service Unavailable, with a text message that informs that this server has the provenance disabled in it.
- Provenance can be enabled on one server but disabled on another.
- Restarting a server with provenance switched from on to off will leave the provenance store intact, but no more data will be committed to it.
- Restarting a server with provenance switched from off to on is not supported  and will result in an error.
    
Some ledger queries that previously depended on the provenance store are now decoupled from the provenance store and continue to be served.


internal/provenance/open.go
```diff
@@ -71,6 +81,15 @@ func Open(conf *Config) (*Store, error) {
 		return openNewProvenanceStore(conf)
 	}
 
+	if conf.Disabled {
+		conf.Logger.Info("The provenance store is disabled on this node.")
+		disabledFlagPath := filepath.Join(conf.StoreDir, disabledFlag)
+		if _, err := os.Create(disabledFlagPath); err != nil {
```
cendhu, line 87: use `fileops.CreateFile()`

cendhu (commented):
One comment. Otherwise, LGTM

internal/provenance/open.go
tock-ibm: The `fileops.CreateFile()` version has semantics of "create if does not exist", whereas the `os.Create()` is "create or truncate". I actually need the later version; since the file is just a flag, it saves me the need to check existence and create only if it does not exist.

cendhu (approved):

PR #457, opened 2022-11-10
Metric features

liran-funaro:
Made the following changes:

 - Add Prometheus network configuration
 - Add block size (bytes) metric (changed previous `block_size` to `tx_per_block`)
 - Add explicit buckets values
 - Add an option to issue certificate using URL (not IP) to the testing infrastructure (used by the benchmark)


internal/utils/stats.go
```diff
@@ -20,156 +49,194 @@ type blockProcessorStats struct {
...
-	s.blockSize.Observe(float64(size))
+func (s *BlockProcessorStats) UpdateBlockSizeBytes(size int64) {
+	s.blockSizeBytes.Observe(float64(size))
 }
+
+func (s *BlockProcessorStats) IncrementTransactionCount(flag types.Flag, tx_type string) {
+	s.transactionCount.WithLabelValues(flag.String(), tx_type).Inc()
+}
+
+var Stats = newBlockProcessorStats()
```
tock-ibm, line 242: This is not a good approach, since we often run multiple instances of components in the same process. See replication, for example. Let's not use global variables but only an instance per component, as Senthil started implementing. It is OK for the performance evaluation, but not for production.

liran-funaro: I generally agree.

Note that Senthil implementation also contained a global stats object. This object was later assigned to each component. But it was still the same global object.
I simply exposed it even further to support the sig validation metric.

I'm not sure what Prometheus does if you define the same metric name twice.
So it is best to only create each Prometheus metric once (globally) and use it everywhere.
Prometheus handles that well.

We could define each metric in the appropriate module, instead of in a general module.
This will be a better code design, but in terms of correctness, this would not make any difference.

PR #459, opened 2022-11-13
Disable MP-Trie commit boot only

tock-ibm:


cendhu:
It would have been good if the check for enablement is contained within mp-trie package but I understood why we have check outside -- mainly due to the existing way of implementation.

Otherwise, LGTM. Please rebase.

cendhu (approved):

Issue #455, opened 2022-11-09
Access control on reading user records during a query


tock-ibm:
Every user, even regular users, can GET the user record of every other user. 

Close the gap by allowing only admins to query user records when ACLs are nil.

Admins should be able to query even if ACLs are not nil regardless of the Readers set.

The Writer set & policy is ignored (we assume it to be empty). Writer must always be an admin.

This issue pertains only to the read (query) path, we'll address the write (TX) path in another issue #456 .

PR #460, opened 2022-11-14
Access control on reading user records during a query

MayRosenbaum:
#455

### DIFF
--- a/internal/bcdb/worldstate_query_processor_test.go
+++ b/internal/bcdb/worldstate_query_processor_test.go
@@ -1302,56 +1302,64 @@ func TestExecuteJSONQuery(t *testing.T) {
 }
 
 func TestGetUser(t *testing.T) {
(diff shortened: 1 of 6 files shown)


internal/bcdb/worldstate_query_processor_test.go
```diff
@@ -1388,19 +1384,19 @@ func TestGetUser(t *testing.T) {
 				},
 			},
 			{
-				name: "querierUser has read-write permission on targetUser",
+				name: "admin always has read permission on targetUser",
```
tock-ibm, line 1387: this is the same as the one below it, I think maybe  here you should change the target user:

`Metadata: targetUserMetadataNoACL,` to  `Metadata: targetUserMetadataReadPerm,`

This way you check admin access without ACL and with ACL.Read existing

tock-ibm (approved):

Issue #456, opened 2022-11-09
Access control on writing user records - restrictive


tock-ibm:
Admins are the only users that can write a user record.
We progress in 2 stages:

1. Allow regular users to be included in the acl.readers set, in order to allow a regular user to read another user's record (for the certificate).
2. No one can be included in the acl.writers set, it must be empty.
3. Admin users can always read/write/delete user records, regardless of ACLs
4. An admin user can not be altered during a UserTx, only during a ConfigTx

PR #467, opened 2022-11-22
Access control on writing user records - restrictive

MayRosenbaum:
Signed-off-by: May Rosenbaum <mayro1595@gmail.com> #456


tock-ibm:
Implemented in #470

PR #469, opened 2022-11-24
Remove redundant lock

liran-funaro:
Using locks causes a long hold of the `transactionProcessor` lock, which is the single contention point of the TX processing.
It can be avoided because the `blockReplicator` uses its own lock and by making `pendingTxs` lock-free.

### DIFF
--- a/internal/bcdb/transaction_processor.go
+++ b/internal/bcdb/transaction_processor.go
@@ -1,11 +1,11 @@
 // Copyright IBM Corp. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+
 package bcdb
 
 import (
 	"encoding/json"
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/google/uuid"
@@ -45,7 +45,6 @@ type transactionProcessor struct {
 	blockStore           *blockstore.Store
 	pendingTxs           *queue.PendingTxs
 	logger               *logger.SugarLogger
-	sync.Mutex
 }
 
 type txProcessorConfig struct {
@@ -266,36 +265,35 @@ func (t *transactionProcessor) SubmitTransaction(tx interface{}, timeout time.Du
 		return nil, err
 	}
 
-	t.Lock()
-	duplicate, err := t.isTxIDDuplicate(txID)
-	if err != nil {
-		t.Unlock()
-		return nil, err
-	}
-	if duplicate {
-		t.Unlock()
+	// We attempt to insert the txID atomically.
+	// If we succeed, then future TX fill fail at this point.
+	// However, if the TX already exists in the block store, then we will fail the subsequent check.
+	// Since a TX will be removed from the pending queue only after it is inserted to the block store,
+	// Then it is guaranteed that we won't use the same txID twice.
+	// TODO: add limit on the number of pending sync tx
+	promise := queue.NewCompletionPromise(timeout)
+	if existed := t.pendingTxs.Add(txID, promise); existed {
 		return nil, &internalerror.DuplicateTxIDError{TxID: txID}
 	}
 
-	if t.txQueue.IsFull() {
-		t.Unlock()
-		return nil, fmt.Errorf("transaction queue is full. It means the server load is high. Try after sometime")
+	duplicate, err := t.blockStore.DoesTxIDExist(txID)
+	if err != nil || duplicate {
+		t.pendingTxs.DeleteWithNoAction(txID)
+		if err == nil {
+			err = &internalerror.DuplicateTxIDError{TxID: txID}
+		}
+		return nil, err
 	}
 
 	jsonBytes, err := json.MarshalIndent(tx, "", "\t")
 	if err != nil {
-		t.Unlock()
 		return nil, fmt.Errorf("failed to marshal transaction: %v", err)
 	}
 	t.logger.Debugf("enqueuing transaction %s\n", string(jsonBytes))
 
+	// Enqueue will block until the queue is not full
 	t.txQueue.Enqueue(tx)
 	t.logger.Debug("transaction is enqueued for re-ordering")
-
-	promise := queue.NewCompletionPromise(timeout)
-	// TODO: add limit on the number of pending sync tx
-	t.pendingTxs.Add(txID, promise)
-	t.Unlock()
 
 	receipt, err := promise.Wait()
 
@@ -341,25 +339,10 @@ func (t *transactionProcessor) PostBlockCommitProcessing(block *types.Block) err
 	return nil
 }
 
-func (t *transactionProcessor) isTxIDDuplicate(txID string) (bool, error) {
-	if t.pendingTxs.Has(txID) {
-		return true, nil
-	}
-
-	isTxIDAlreadyCommitted, err := t.blockStore.DoesTxIDExist(txID)
-	if err != nil {
-		return false, err
-	}
-	return isTxIDAlreadyCommitted, nil
-}
-
 func (t *transactionProcessor) Close() error {
-	t.Lock()
-	defer t.Unlock()
-
 	t.txReorderer.Stop()
 	t.blockCreator.Stop()
-	t.blockReplicator.Close()
+	_ = t.blockReplicator.Close()
 	t.peerTransport.Close()
 	t.blockProcessor.Stop()
 
@@ -373,9 +356,6 @@ func (t *transactionProcessor) IsLeader() *internalerror.NotLeaderError {
 // ClusterStatus returns the leader NodeID, and the active nodes NodeIDs.
 // Note: leader is always in active.
 func (t *transactionProcessor) ClusterStatus() (leader string, active []string) {
-	t.Lock()
-	defer t.Unlock()
-
 	leaderID, activePeers := t.blockReplicator.GetClusterStatus()
 	for _, peer := range activePeers {
 		active = append(active, peer.NodeId)
--- a/internal/bcdb/transaction_processor.go
+++ b/internal/bcdb/transaction_processor.go
@@ -5,7 +5,6 @@ package bcdb
 
 import (
 	"encoding/json"
-	"fmt"
 	"time"
 
 	"github.com/google/uuid"
@@ -269,7 +268,7 @@ func (t *transactionProcessor) SubmitTransaction(tx interface{}, timeout time.Du
 	// If we succeed, then future TX fill fail at this point.
 	// However, if the TX already exists in the block store, then we will fail the subsequent check.
 	// Since a TX will be removed from the pending queue only after it is inserted to the block store,
-	// Then it is guaranteed that we won't use the same txID twice.
+	// then it is guaranteed that we won't use the same txID twice.
 	// TODO: add limit on the number of pending sync tx
 	promise := queue.NewCompletionPromise(timeout)
 	if existed := t.pendingTxs.Add(txID, promise); existed {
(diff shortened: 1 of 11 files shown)


internal/bcdb/transaction_processor.go
```diff
@@ -266,37 +265,36 @@ func (t *transactionProcessor) SubmitTransaction(tx interface{}, timeout time.Du
...
-	if duplicate {
-		t.Unlock()
+	// We attempt to insert the txID atomically.
+	// If we succeed, then future TX fill fail at this point.
+	// However, if the TX already exists in the block store, then we will fail the subsequent check.
+	// Since a TX will be removed from the pending queue only after it is inserted to the block store,
+	// Then it is guaranteed that we won't use the same txID twice.
+	// TODO: add limit on the number of pending sync tx
+	promise := queue.NewCompletionPromise(timeout)
+	if existed := t.pendingTxs.Add(txID, promise); existed {
```
cendhu, line 275: Why did you remove this part of the code?

```
	if t.pendingTxs.Has(txID) {
 		return true, nil
 	}
```

We want to ensure that there is no duplicate txID either in block store or provenance store. Hence, we reject any transaction that has a duplicate txID with either pendingTxs or block store. At all time, our block store has unique txIDs. Whoever we accept a transaction, we check whether the txID is duplicate by looking at all pending txs and committed states. Only if it is found to be unique, we would accept the transaction. Otherwise, it is rejected.

liran-funaro: `t.pendingTxs.Add(txID, promise)` will return `existed=true` if it was already in the map.
So this one atomic call replaces the two non-atomic calls, avoiding the [TocTou](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issue.

cendhu: got it. thanks.

internal/bcdb/transaction_processor.go
```diff
@@ -266,37 +265,36 @@ func (t *transactionProcessor) SubmitTransaction(tx interface{}, timeout time.Du
...
 
 	jsonBytes, err := json.MarshalIndent(tx, "", "\t")
 	if err != nil {
-		t.Unlock()
 		return nil, fmt.Errorf("failed to marshal transaction: %v", err)
 	}
 	t.logger.Debugf("enqueuing transaction %s\n", string(jsonBytes))
 
+	// Enqueue will block until the queue is not full
 	t.txQueue.Enqueue(tx)
```
liran-funaro, line 295: The enqueue should respect the request timeout

pkg/logger/logger.go
```diff
@@ -76,7 +76,12 @@ func (l *SugarLogger) With(args ...interface{}) *SugarLogger {
 	}
 }
 
-func (l *SugarLogger) SetLogLevel(level string) error {
+func (l *SugarLogger) IsDebug() bool {
+	return l.conf.Level.Level() == zap.DebugLevel
+}
+
+// setLogLevel is used only for tests
+func (l *SugarLogger) setLogLevel(level string) error {
```
tock-ibm, line 84: For now, it is used only for tests, but we might add operation endpoints for changing it dynamically at runtime, so lets keep it public.

internal/queue/pending_txs.go
```diff
@@ -57,27 +64,25 @@ func (p *PendingTxs) DoneWithReceipt(txIDs []string, blockHeader *types.BlockHea
...
-	p.RLock()
-	defer p.RUnlock()
-
-	return len(p.txs) == 0
+	empty := true
+	p.txs.Range(func(key, value interface{}) bool {
+		empty = false
+		return false
+	})
+	return empty
```
tock-ibm, line 87: This has semantics different than the previous Empty(), as Range does not respect any snapshot. It is probably OK to use in tests, but not in production (we only use in tests, lets keep it this way). Maybe we should put a comment on the method in line with the comment on Range.

internal/bcdb/transaction_processor.go
```diff
@@ -266,37 +264,47 @@ func (t *transactionProcessor) SubmitTransaction(tx interface{}, timeout time.Du
...
-		return nil, err
-	}
-	if duplicate {
-		t.Unlock()
+	// We attempt to insert the txID atomically.
+	// If we succeed, then future TX fill fail at this point.
+	// However, if the TX already exists in the block store, then we will fail the subsequent check.
+	// Since a TX will be removed from the pending queue only after it is inserted to the block store,
+	// then it is guaranteed that we won't use the same txID twice.
+	// TODO: add limit on the number of pending sync tx
```
tock-ibm, line 272: Placing a limit will be a problem with sync.Map

internal/blockstore/commit_and_query.go
```diff
@@ -477,9 +477,6 @@ func (s *Store) GetHeaderByHash(blockHash []byte) (*types.BlockHeader, error) {
 // DoesTxIDExist returns true if any of the committed block has a transaction with
 // the given txID. Otherwise, it returns false
 func (s *Store) DoesTxIDExist(txID string) (bool, error) {
-	s.mu.RLock()
-	defer s.mu.RUnlock()
-
```
tock-ibm, line 482: It is correct to remove this lock only because this method is used optimistically from the front end. Lets ass a comment that specifies that, such that it is never used in the validation & commit path.

pkg/logger/logger.go
```diff
@@ -76,6 +76,11 @@ func (l *SugarLogger) With(args ...interface{}) *SugarLogger {
 	}
 }
 
+func (l *SugarLogger) IsDebug() bool {
+	return l.conf.Level.Level() == zap.DebugLevel
+}
```
tock-ibm, line 81: why no mutex.Lock()?

tock-ibm: Or maybe, after looking into the SugarLogger, why mutex at all? as the levels are atomic?
Plus, the underlying logger has `Enabled(level)`, which seems like the right thing to use.

internal/blockstore/commit_and_query.go
liran-funaro: It is safe to remove the locks because this method calls a function of one of its internal classes, which is protected.

internal/queue/pending_txs.go
```diff
@@ -11,43 +11,50 @@ import (
 )
 
 type PendingTxs struct {
-	sync.RWMutex
-	txs map[string]*CompletionPromise
-
+	txs    sync.Map
```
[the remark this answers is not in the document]
liran-funaro: Indeed, a simple micro-benchmark that adds/deletes values from the map (50:50) shows a 25% speedup with a simple `sync.Mutex` instead of `sync.Map`.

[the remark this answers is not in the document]
liran-funaro: Reverted back to map

internal/queue/pending_txs.go
liran-funaro: Outdated. Reverted back to map.

internal/bcdb/transaction_processor.go
liran-funaro: Outdated. Reverted back to map.

pkg/logger/logger.go
liran-funaro: Changed to use the suggested API.

internal/queue/pending_txs.go
```diff
@@ -24,30 +23,49 @@ func NewPendingTxs(logger *logger.SugarLogger) *PendingTxs {
...
-	p.Lock()
-	defer p.Unlock()
-
 	for txIndex, txID := range txIDs {
-		p.txs[txID].done(
+		promise, loaded := p.loadAndDelete(txID)
+		if !loaded {
+			continue
+		}
+		promise.done(
```
tock-ibm, line 63: Does it make sense to take the lock len(txIDs) times instead of once? 
We could also `loadAndDelete` all the promises into a slice `[]promise` (known size, can pre-allocate)  under the lock and then call `done` in a loop out side the lock.  
We want to minimize the total time the lock is taken and the number of times it is taken.

internal/queue/pending_txs.go
```diff
@@ -24,30 +23,49 @@ func NewPendingTxs(logger *logger.SugarLogger) *PendingTxs {
...
+}
 
-	p.txs[txID] = promise
+func (p *PendingTxs) loadAndDelete(txID string) (*CompletionPromise, bool) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	promise, loaded := p.txs[txID]
+	if loaded {
+		delete(p.txs, txID)
+	}
```
tock-ibm, line 49: Most of the time `loaded=true`. It is `false` only when a node is committing after a leadership change with pending TXs in the Raft queues which are not in the PendindTxs queue, as they were submitted in another node (which shutdown). This is extremely rare, so we might as well just delete without checking.

internal/queue/pending_txs.go
liran-funaro: There is a tradeoff here. Locking once will improve the commit time but will slow down the insert-to-queue time.
Since the number of TXs can be 1e3 up to 1e6, I'm not sure what is the best approach.
A way to avoid the commit time wasted on locking will be to invoke the promises on a non-blocking go routine.
I.e., call `internal/blockprocessor/processor.go:136` inside a go routine without waiting for it.

tock-ibm: We definitely want to reduce the commit time. 
So I think that is a good idea - delegate the release of the promises to a goroutine. I still think that goroutine should take the lock once for every block, not once for every tx. This will reduce contention for the lock and will reduce the CPU overall.

liran-funaro: Following is a micro-benchmark of 50:50 insert-if-absent/delete-if-present. The delete-if-present operations are batched, i.e., multiple keys are deleted in the same go routine, while each insert-if-absent is executed in its own go routine.

I compared four implementations: `sync.map`, [`xsync.map`](https://github.com/puzpuzpuz/xsync), regular map with a mutex, and regular map with a mutex that is locked only once per delete batch.

The X axis is the size of the delete-batch, i.e., how many keys are deleted in a single batch.
The Y axis is the operation time (nanoseconds).

This shows that batch locking can reduce the operation's time, but the use of `xsync` can improve the performance significantly.

image

internal/blockprocessor/processor.go
```diff
@@ -133,9 +134,11 @@ func (b *BlockProcessor) Start() {
...
 			}
 
-			if err = b.listeners.invoke(block); err != nil {
-				panic(err)
-			}
+			go func(curBlock *types.Block) {
+				if err = b.listeners.invoke(curBlock); err != nil {
+					panic(err)
+				}
+			}(block)
```
tock-ibm, line 141: You cant start the goroutine here, those are general listeners that not all will want to behave async like that. I think it is better start it in the PendingTx.DoneWithReceipt(...) method.

internal/queue/pending_txs_test.go
```diff
@@ -4,14 +4,15 @@
...
 
 import (
-	ierrors "github.com/hyperledger-labs/orion-server/internal/errors"
 	"sync"
 	"testing"
 	"time"
 
+	ierrors "github.com/hyperledger-labs/orion-server/internal/errors"
+
+	"github.com/golang/protobuf/proto"
```
tock-ibm, line 13: goimports

internal/blockprocessor/processor.go
liran-funaro: I moved the go routine invocation to `transaction_processor.go`.
FYI, currently, there is only one listener implementation in the code, and this is the `transactionProcessor` which invokes the promises' `DoneWithReceipt()`.

tock-ibm (approved):

PR #477, opened 2022-12-25
Encode keys in URLs with base64url raw

tock-ibm:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,9 @@ cmd/bdb/bdb
 
 #IDE artifacts
 .idea
+
+#Documentation with docusaurus
+.docusaurus
+node_modules
+yarn.lock
+
--- a/internal/httphandler/data_request_handler.go
+++ b/internal/httphandler/data_request_handler.go
@@ -1,5 +1,6 @@
 // Copyright IBM Corp. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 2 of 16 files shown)


liran-funaro (dismissed):
My only note is that the issue that this PR solves can appear for every string that is passed via a rest API. This includes the keys, but also the DB name and the user ID.
I suggest modifying the API to support base64 encoding for these fields.
Otherwise, a user might be able to create a field (via post request) that it cannot access.
Alternatively, we can add a separate rest API (e.g., with a prefix /base64/...) that all names are referred to using a base64 encoding (in parallel to the plain text API).

tock-ibm:
> My only note is that the issue that this PR solves can appear for every string that is passed via a rest API. This includes the keys, but also the DB name and the user ID. I suggest modifying the API to support base64 encoding for these fields. Otherwise, a user might be able to create a field (via post request) that it cannot access. Alternatively, we can add a separate rest API (e.g., with a prefix /base64/...) that all names are referred to using a base64 encoding (in parallel to the plain text API).

I agree that this is a potential problem to every field, and this is indicated in the issue. In other field we can enforce naming conventions that are URL and file friendly, and rejecting TXs that try and create usernames and database names that violate these restrictions, e.g. https://github.com/hyperledger-labs/orion-server/issues/49

liran-funaro (approved):

PR #481, opened 2022-12-31
test: use `T.TempDir` to create temporary test directory

Juneezee:
A testing cleanup. 

This pull request replaces `ioutil.TempDir` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := ioutil.TempDir("", "")
	require.NoError(t, err)
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

### DIFF
--- a/internal/fileops/file_ops_test.go
+++ b/internal/fileops/file_ops_test.go
@@ -371,26 +371,15 @@ func TestSyncDir(t *testing.T) {
 func prepareTestDir(t *testing.T) string {
 	tempDir := t.TempDir()
 
-	if err := os.Mkdir(path.Join(tempDir, "dir"), 0755); err != nil {
-		t.FailNow()
-	}
-	if err := os.Mkdir(path.Join(tempDir, "dir", "a"), 0755); err != nil {
-		t.FailNow()
-	}
-	if err := os.Mkdir(path.Join(tempDir, "dir", "b"), 0755); err != nil {
-		t.FailNow()
-	}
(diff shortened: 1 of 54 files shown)


internal/fileops/file_ops_test.go
```diff
@@ -380,31 +369,25 @@ func TestSyncDir(t *testing.T) {
...
-		removeAndFailNow()
+		t.FailNow()
 	}
 	if err := os.Mkdir(path.Join(tempDir, "dir", "c"), 0755); err != nil {
-		removeAndFailNow()
+		t.FailNow()
 	}
 	if err := os.Mkdir(path.Join(tempDir, "dir", "d"), 0755); err != nil {
-		removeAndFailNow()
+		t.FailNow()
```
liran-funaro, line 387: Can we use here `require.NoError(t, ...)` instead of `if err := ...; err != nil {t.FailNow()}`?

liran-funaro (approved):
Thanks!

internal/fileops/file_ops_test.go
Juneezee: Sure :ok_hand:

liran-funaro (approved):
Great!

PR #488, opened 2023-01-23
Metrics

liran-funaro:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -31,6 +31,7 @@ type LocalConfiguration struct {
 	BlockCreation BlockCreationConf
 	Replication   ReplicationConf
 	Bootstrap     BootstrapConf
+	Prometheus    PrometheusConf
 }
 
 // ReplicationConf provides local configuration parameters for replication and server to server communication.
@@ -77,7 +78,7 @@ type ServerConf struct {
 	// The database configuration of the local node.
 	Database DatabaseConf
 	// The provenance store configuration of the local node.
-	Provenance    ProvenanceConf
+	Provenance ProvenanceConf
 	// The lengths of various queues that buffer between internal components.
 	QueueLength QueueLengthConf
 	// QueryProcessing holds limits associated with query responses
@@ -160,6 +161,12 @@ type BootstrapConf struct {
 	// File contains the path to initial configuration that will be used to bootstrap the node,
 	// as specified by the`Method`.
 	File string
+}
+
+type PrometheusConf struct {
+	Enabled bool
+	Network NetworkConf
+	TLS     TLSConf
 }
 
 // Read reads configurations from the config file and returns the config
--- a/config/config.go
+++ b/config/config.go
@@ -163,10 +163,14 @@ type BootstrapConf struct {
 	File string
 }
 
+// PrometheusConf specifies the metrics collection for monitoring and performance analysis.
 type PrometheusConf struct {
+	// Enabled specifies if metrics will be collected
 	Enabled bool
+	// Network defines the listen address and port used for the Prometheus server.
 	Network NetworkConf
-	TLS     TLSConf
+	// TLS defines TLS settings for the Prometheus server.
+	TLS TLSConf
 }
 
 // Read reads configurations from the config file and returns the config
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -88,8 +88,6 @@ var expectedLocalConfig = &LocalConfiguration{
 			ClientAuthRequired:    false,
 			ServerCertificatePath: "./testdata/cluster/server.cert",
 			ServerKeyPath:         "./testdata/cluster/server.key",
-			ClientCertificatePath: "./testdata/cluster/client.cert",
-			ClientKeyPath:         "./testdata/cluster/client.key",
 			CaConfig: CAConfiguration{
 				RootCACertsPath:         []string{"./testdata/cluster/rootca.cert"},
 				IntermediateCACertsPath: []string{"./testdata/cluster/midca.cert"},
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -77,6 +77,25 @@ var expectedLocalConfig = &LocalConfiguration{
 		Method: "genesis",
 		File:   "./testdata/3node-shared-config-bootstrap.yml",
 	},
+	Prometheus: PrometheusConf{
+		Enabled: true,
+		Network: NetworkConf{
+			Address: "127.0.0.1",
+			Port:    8050,
+		},
+		TLS: TLSConf{
+			Enabled:               false,
+			ClientAuthRequired:    false,
+			ServerCertificatePath: "./testdata/cluster/server.cert",
+			ServerKeyPath:         "./testdata/cluster/server.key",
+			ClientCertificatePath: "./testdata/cluster/client.cert",
+			ClientKeyPath:         "./testdata/cluster/client.key",
+			CaConfig: CAConfiguration{
+				RootCACertsPath:         []string{"./testdata/cluster/rootca.cert"},
+				IntermediateCACertsPath: []string{"./testdata/cluster/midca.cert"},
+			},
+		},
+	},
 }
 
 func TestConfig(t *testing.T) {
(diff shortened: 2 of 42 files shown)


config/config.go
```diff
@@ -162,6 +163,12 @@ type BootstrapConf struct {
 	File string
 }
 
+type PrometheusConf struct {
+	Enabled bool
+	Network NetworkConf
+	TLS     TLSConf
+}
+
```
tock-ibm, line 171: looks good. please document it and add some tests to config_test.go and items to the testdata folder that would cover the correct reading of a config with this item.

internal/httphandler/config_request_handler.go
```diff
@@ -29,7 +30,7 @@ type configRequestHandler struct {
 }
 
 // NewConfigRequestHandler return config query and transactions request handler
-func NewConfigRequestHandler(db bcdb.DB, logger *logger.SugarLogger) http.Handler {
+func NewConfigRequestHandler(db bcdb.DB, logger *logger.SugarLogger, metricsRegistry *prometheus.Registry) http.Handler {
```
tock-ibm, line 33: That is relevant to all handlers - we should probably start with data and move on from there.

internal/bcdb/transaction_processor.go
```diff
@@ -304,13 +366,16 @@ func (t *transactionProcessor) SubmitTransaction(tx interface{}, timeout time.Du
 		}
 	}
 	t.logger.Debug("transaction is enqueued for re-ordering")
+	t.metrics.QueueSize("tx", t.txQueue.Size())
+	t.metrics.QueueSize("pending", t.pendingTxs.Size())
```
tock-ibm, line 370: Lets use these two metrics to test the metrics endpoint server from the out side. once we lay the infrastructure for it, and test it, we can easily add metrics to additional units following the same pattern (i.e. handlers, etc).

pkg/server/server.go
```diff
@@ -163,6 +207,28 @@ func (s *BCDBHTTPServer) serveRequests(l net.Listener) {
 	s.logger.Infof("Finished serving requests on: %s", s.listen.Addr().String())
 }
 
+func (s *BCDBHTTPServer) servePrometheus() {
+	if s.metricsServer == nil {
+		return
+	}
```
tock-ibm, line 213: isn't it better to do this around the the `go s.servePrometheus()`?

pkg/server/server.go
```diff
@@ -44,18 +48,46 @@ func New(conf *config.Configurations) (*BCDBHTTPServer, error) {
 		return nil, err
 	}
 
-	db, err := bcdb.NewDB(conf, lg)
+	var metricsRegistry *prometheus.Registry
+	var metricsServer *http.Server
+	var metricsListen net.Listener
```
tock-ibm, line 53: Lets add a test that checks the endpoint can be used with one or two metrics, like queue size, etc, before moving on

pkg/server/server.go
```diff
@@ -44,18 +48,46 @@ func New(conf *config.Configurations) (*BCDBHTTPServer, error) {
...
+	if conf.LocalConfig.Prometheus.Enabled {
+		metricsRegistry = prometheus.NewRegistry()
+		metricsRegistry.MustRegister(
+			prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}),
+			prometheus.NewGoCollector(),
+		)
```
tock-ibm, line 59: how do units that get the registry behave if it is nil? do they still collect metrics? i.e. counters, measure time, etc?

internal/bcdb/transaction_processor.go
```diff
@@ -25,12 +26,67 @@ import (
...
 
 const (
 	commitListenerName = "transactionProcessor"
 )
 
+type transactionProcessorMetrics struct {
+	enabled    bool
+	queueSize  *prometheus.GaugeVec
+	submitTime prometheus.Histogram
+}
```
tock-ibm, line 40: put this is a different go file and test it separately.

config/config.go
liran-funaro: added

pkg/server/server.go
liran-funaro: They are both equally good, IMO. What do you prefer?

liran-funaro:
Largest projects that use Prometheus go client:
| project                                        | stars   | use |
|--------------------------------------------|---------|--|
| [ethereum/go-ethereum](https://github.com/ethereum/go-ethereum)    | 41K     | indirect via [deathowl/go-metrics-prometheus](https://github.com/deathowl/go-metrics-prometheus) |
| [grafana/k6](https://github.com/grafana/k6)              | 20K     | indirect |
| [cloudreve/Cloudreve](https://github.com/cloudreve/Cloudreve)     | 17K     | indirect |
| [redis/go-redis](https://github.com/redis/go-redis)          | 17K     | single collector periodically monitoring resources |
| [aquasecurity/trivy](https://github.com/aquasecurity/trivy)       | 16K     | indirect |
| [authelia/authelia](https://github.com/authelia/authelia)        | 16K     | single metrics object |
| [kubernetes/dashboard](https://github.com/kubernetes/dashboard)     | 12K     | single collector periodically monitoring resources |
| [apple/foundationdb](https://github.com/apple/foundationdb)       | 12K     | single collector periodically monitoring resources |
| [projectdiscovery/nuclei](https://github.com/projectdiscovery/nuclei)  | 12K     | indirect |
| [goreleaser/goreleaser](https://github.com/goreleaser/goreleaser)    | 11K     | indirect |

[deathowl/go-metrics-prometheus](https://github.com/deathowl/go-metrics-prometheus) uses single metrics object with a map of dynamically defined metrics.

pkg/server/server.go
liran-funaro: When the registry is nil, the metrics object methods do nothing.

test/setup/server_setup.go
```diff
@@ -1476,6 +1496,13 @@ func (s *Server) URL() string {
 	return "http://" + s.address + ":" + strconv.FormatInt(int64(s.nodePort), 10)
 }
 
+func (s *Server) PrometheusURL() string {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	return "http://" + s.address + ":" + strconv.FormatInt(int64(s.prometheusPort), 10)
+}
+
```
tock-ibm, line 1505: do we have a test with TLS? the scheme should be http/https depending on whether security is on

internal/utils/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+	reg.MustRegister(s.txSize)
+	return s
+}
+
+func (s *DataRequestHandlingMetrics) Latency(label string, startTime time.Time) {
+	// We redefine this method to prevent segmentation fault
+	if s != nil {
+		s.commonMetrics.Latency(label, startTime)
+	}
+}
```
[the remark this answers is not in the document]
liran-funaro: For some reason, when accessing a parent class method, it tries to dereference the object.
It seems weird to me because the correct method call can be resolved at compile time, but it is what it is.
So I resolved this by redefining the method in the child class.

config/config_test.go
```diff
@@ -77,6 +77,25 @@ var expectedLocalConfig = &LocalConfiguration{
...
+			Address: "127.0.0.1",
+			Port:    8050,
+		},
+		TLS: TLSConf{
+			Enabled:               false,
+			ClientAuthRequired:    false,
+			ServerCertificatePath: "./testdata/cluster/server.cert",
+			ServerKeyPath:         "./testdata/cluster/server.key",
+			ClientCertificatePath: "./testdata/cluster/client.cert",
+			ClientKeyPath:         "./testdata/cluster/client.key",
```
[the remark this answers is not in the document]
liran-funaro: We discussed before about client-auth for metrics and we realized the TLS client cert validation acts as client auth.

test/setup/server_setup.go
liran-funaro: I've added such test

tock-ibm (approved):

PR #490, opened 2023-01-30
Cache LevelDB snapshots and use lock-free map

liran-funaro:
- Replacing map with sync.Map
- Caching DB snapshot after commit to avoid creating a snapshot for each Get operation
- Modifying cache to return byte array to avoid code duplication
- Remove Db lock because levelDB is already protected
- Reduce code duplication in Open()

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/syndtr/goleveldb v1.0.1-0.20220721030215-126854af5e6d
 	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2
 	go.etcd.io/etcd v0.5.0-alpha.5.0.20210226220824-aa7126864d82 // indirect git tag v3.4.15
+	go.uber.org/multierr v1.6.0
 	go.uber.org/zap v1.18.1
 	google.golang.org/protobuf v1.26.0
 	gopkg.in/yaml.v2 v2.4.0
--- a/internal/worldstate/leveldb/cache.go
+++ b/internal/worldstate/leveldb/cache.go
@@ -2,8 +2,6 @@ package leveldb
 
 import (
 	"github.com/VictoriaMetrics/fastcache"
-	"github.com/golang/protobuf/proto"
-	"github.com/hyperledger-labs/orion-server/pkg/types"
 )
 
 var keySep = []byte{0x00}
@@ -27,19 +25,9 @@ func newCache(dataCacheSizeMBs int) *cache {
 
 // getState returns the value for a given namespace and key from
 // the cache
-func (c *cache) getState(namespace, key string) (*types.ValueWithMetadata, error) {
+func (c *cache) getState(namespace, key string) ([]byte, bool) {
 	cacheKey := constructCacheKey(namespace, key)
-
-	valBytes, exists := c.dataCache.HasGet(nil, cacheKey)
-	if !exists {
-		return nil, nil
-	}
-
-	cacheValue := &types.ValueWithMetadata{}
-	if err := proto.Unmarshal(valBytes, cacheValue); err != nil {
-		return nil, err
-	}
-	return cacheValue, nil
+	return c.dataCache.HasGet(nil, cacheKey)
 }
 
 // putState stores a given value in the cache
--- a/internal/worldstate/leveldb/cache_test.go
+++ b/internal/worldstate/leveldb/cache_test.go
@@ -9,6 +9,17 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+func unmarshaler(t *testing.T) func(b []byte, inCache bool) (*types.ValueWithMetadata, bool) {
+	return func(b []byte, inCache bool) (*types.ValueWithMetadata, bool) {
+		if !inCache {
+			return nil, inCache
+		}
+		value := &types.ValueWithMetadata{}
+		require.NoError(t, proto.Unmarshal(b, value))
+		return value, inCache
+	}
+}
+
 func TestCache(t *testing.T) {
 	cache := newCache(10)
 
@@ -34,8 +45,8 @@ func TestCache(t *testing.T) {
 
 	t.Run("check basic storage and retrieval", func(t *testing.T) {
 		// db1, key1 does not exist
-		v, err := cache.getState("db1", "key1")
-		require.NoError(t, err)
+		v, inCache := cache.getState("db1", "key1")
+		require.False(t, inCache)
 		require.Nil(t, v)
 
 		s := &fastcache.Stats{}
@@ -54,11 +65,12 @@ func TestCache(t *testing.T) {
 		}
 		valBytes, err := proto.Marshal(db1Key1Value1)
 		require.NoError(t, err)
-		cache.putState("db1", "key1", valBytes)
+		require.NoError(t, cache.putState("db1", "key1", valBytes))
 
-		// db1, key1 should exists
-		actualKey1Value1, err := cache.getState("db1", "key1")
-		require.NoError(t, err)
+		// db1, key1 should exist
+		actualKey1Value1, inCache := unmarshaler(t)(cache.getState("db1", "key1"))
+		require.True(t, inCache)
+		require.NotNil(t, actualKey1Value1)
 		require.True(t, proto.Equal(db1Key1Value1, actualKey1Value1))
 
 		cache.dataCache.UpdateStats(s)
@@ -69,11 +81,12 @@ func TestCache(t *testing.T) {
 		// update key1's value
 		valBytes, err := proto.Marshal(db1Key1Value2)
 		require.NoError(t, err)
-		cache.putState("db1", "key1", valBytes)
+		require.NoError(t, cache.putState("db1", "key1", valBytes))
 
 		// db1, key1 should have the updated value
-		actualKey1Value2, err := cache.getState("db1", "key1")
-		require.NoError(t, err)
+		actualKey1Value2, inCache := unmarshaler(t)(cache.getState("db1", "key1"))
+		require.True(t, inCache)
+		require.NotNil(t, actualKey1Value2)
 		require.True(t, proto.Equal(db1Key1Value2, actualKey1Value2))
 
 		s := &fastcache.Stats{}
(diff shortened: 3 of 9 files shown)


cendhu:
The reason for having read lock and write lock between read function and commit function is that when the commit function deletes a DB, it should not result in any panic for the read function. When a DB is deleted, the file is closed and also the directory is removed. In such a time, if a read operation on the db is executed (due a get rest api from the client), server could result in a panic. It all depends on thread scheduling. When deleting a DB, we release the associated snapshot but it is possible for other thread to use it and can cause a panic at the server. 

Check wether the delete operation can be deferred till all read operations complete or revert the changes. If we make committer to wait for some operations to be completed, it would increase the commit time and hence, a degradation in the performance.

liran-funaro:
@cendhu
The scenario you are describing will not result in a panic.
In most cases, it will not even be noticeable.

In POSIX-like OS (like Linux and Mac), when a process holds a file descriptor to a file and a different process/thread deletes that file, the file is not really deleted until the file descriptor is released. It may not appear in the directory anymore, but the inodes still exist and are usable.
So subsequent read operations will not fail and will return valid values.

The only time it will fail is when it will try to create a new file (e.g., a new log file).
Then, it will still not panic. It will simply return an error (could not open file), which will be returned to the user.
This is acceptable behavior since the DB does not exist anymore.

liran-funaro:
@cendhu 
I just noticed that all of the above is irrelevant because we first close the DB, so all operations will simply return the error: "closed".

cendhu:
> @cendhu
> I just noticed that all of the above is irrelevant because we first close the DB, so all operations will simply return the error: "closed".

If it does not result in panic but error, we would return internal server error to the client and it is wrong.

liran-funaro:
I will check in all places when an error is returned to the user, if the error is "closed".
And will return "DB not found" error.
Is this acceptable?

cendhu:
> I will check in all places when an error is returned to the user, if the error is "closed".
> And will return "DB not found" error.
> Is this acceptable?

How much performance improvement are we getting with these changes? If it is justifiable, we can make the changes. Otherwise, it does look like a hacky solution.

cendhu:
> I will check in all places when an error is returned to the user, if the error is "closed".
> And will return "DB not found" error.
> Is this acceptable?

Are we getting the same error with snapshot release too? We could do something like that at the httphandler package.

liran-funaro:
I don't remeber the actual numbers. I can run it again if you want. But I so, I'll do it after we settle on the metrics design.

liran-funaro:
Good thing that you mentioned the Snapshots. It seems there is a bug in the GoLevelDB implementation.
When accessing the released snapshot, it panics instead of returning `leveldb.ErrSnapshotReleased`. 
Looking at their code, they intended to return that error, but they fail on`nil` pointer deference before that.
Unfortunately, GoLevelDB is no longer maintained so I'm not sure how to continue.
We could fork GoLevelDB and fix this issue alone, then we'll import from this fork instead of the original which is not maintained.

liran-funaro:
@cendhu Please correct me if I'm wrong, but we will never delete a DB in the same TX of (or in parallel to) updating it, right?
Because deleting a DB happens in a config TX, which isn't concurrent to any other TX.
So our only concern is when reading from the DB and deleting it concurrently, right?

liran-funaro:
The following demonstrates that the issues we discussed do not occur in our case.

```go
package tests

import (
	"encoding/binary"
	"os"
	"testing"

	"github.com/stretchr/testify/require"
	// Fetched using: `go1.19.4 get github.com/syndtr/goleveldb@HEAD`
	"github.com/syndtr/goleveldb/leveldb" // @HEAD
	"github.com/syndtr/goleveldb/leveldb/iterator"
	"github.com/syndtr/goleveldb/leveldb/opt"
	"github.com/syndtr/goleveldb/leveldb/util"
)

func val(i uint32) (v []byte) {
	v = make([]byte, 4)
	binary.LittleEndian.PutUint32(v, i)
	return
}

func num(v []byte) uint32 {
	return binary.LittleEndian.Uint32(v)
}

func requireReadAll(t *testing.T, reader leveldb.Reader, n uint32) {
	for i := uint32(0); i < n; i++ {
		getValue, err := reader.Get(val(i), nil)
		require.NoError(t, err, "key: %d", i)
		require.Equal(t, val(i+1), getValue, "key: %d", i)
	}
}

func requireIterAll(t *testing.T, it iterator.Iterator, n uint32) {
	m := make(map[uint32]any)

	for i := uint32(0); i < n; i++ {
		require.True(t, it.Next(), "iter: %d", i)
		require.NoError(t, it.Error(), "iter: %d", i)
		k := it.Key()
		v := it.Value()
		require.NotNil(t, k, "iter: %d", i)
		require.NotNil(t, v, "iter: %d", i)
		nk := num(k)
		require.Equal(t, nk+1, num(v))
		m[nk] = nil
	}

	// Check that iterator ends
	require.False(t, it.Next())
	require.NoError(t, it.Error())

	// Check we iterated all keys
	require.Len(t, m, int(n))
}

func Test_LevelDB_Remove(t *testing.T) {
	// Create DB and fill
	f := "test_db"
	require.NoError(t, os.RemoveAll(f))
	file, err := leveldb.OpenFile(f, &opt.Options{})
	require.NoError(t, err)
	require.NotNil(t, file)
	n := uint32(100_000)
	for i := uint32(0); i < n; i++ {
		require.NoError(t, file.Put(val(i), val(i+1), nil), "key: %d", i)
	}

	// Read all (file)
	requireReadAll(t, file, n)

	// Snap
	snap, err := file.GetSnapshot()
	require.NoError(t, err)
	require.NotNil(t, snap)

	// Read all (snapshot)
	requireReadAll(t, snap, n)

	// Create two iterators
	it1 := snap.NewIterator(&util.Range{}, nil)
	require.NoError(t, it1.Error())
	it2 := snap.NewIterator(&util.Range{}, nil)
	require.NoError(t, it2.Error())

	// Release snapshot
	snap.Release()

	// ===========================================================================================
	// Demo 1: Attempt to read a released snapshot returns `leveldb.ErrSnapshotReleased`
	// ===========================================================================================
	_, err = snap.Get(val(0), nil)
	require.IsType(t, leveldb.ErrSnapshotReleased, err)
	// ===========================================================================================

	// ===========================================================================================
	// Demo 2: Attempt to create a new iterator on released snapshot returns `leveldb.ErrSnapshotReleased`
	// ===========================================================================================
	itRelease := snap.NewIterator(&util.Range{}, nil)
	require.IsType(t, leveldb.ErrSnapshotReleased, itRelease.Error())
	// ===========================================================================================

	// ===========================================================================================
	// Demo 3: Attempt to use iterator that was created before the snapshot was released works
	// ===========================================================================================
	requireIterAll(t, it1, n)
	// ===========================================================================================

	// Open another snapshot to test with the closed file
	snap2, err := file.GetSnapshot()
	require.NoError(t, err)
	require.NotNil(t, snap2)

	// Close file and Remove DB files
	require.NoError(t, file.Close())
	require.NoError(t, os.RemoveAll(f))

	// ===========================================================================================
	// Demo 4: Attempt to read an open snapshot of a closed file returns `leveldb.ErrClosed`
	// ===========================================================================================
	_, err = snap2.Get(val(0), nil)
	require.IsType(t, leveldb.ErrClosed, err)
	// ===========================================================================================

	// ===========================================================================================
	// Demo 5: Attempt to use iterator that was created before the snapshot was released
	//         and the DB file was closed and deleted works
	// ===========================================================================================
	requireIterAll(t, it2, n)
	// ===========================================================================================
}

```

cendhu:
what does not occur? Are you saying that no error occurs when the db is closed without releasing snapshots or iterator? 

To be honest, I am not sure about the performance gain that we achieve worth the risk and complexity that we are adding to the code. Especially, for future maintenance, we have to always remember the impact of not having these locks whenever we add need features. If performance gain is significant for the cost we are paying, we can go ahead.

liran-funaro:
Are you suggesting extending `dbList` lock to also protect read/write? Or simply eliminating `mu` lock without changing anything else?

I don't think that assigning two roles to one lock is good practice.
The `dbList` lock is supposed to protect the map, nothing else.
Using it to also protect read/writes will make all reads/writes sequential, even if they are to different DBs.

If we decide to eliminate `mu` lock, and also fix the usage of `dbList` in the `Get()` method to only protect the map, then the changes in this PR are required in order to assure correct execution.

That said, I can shift the snapshot cache modification to a different PR which will be pending future performance evaluation.

cendhu:
> * The `mu` lock does not protect from anything since it is always locked together with the LevelDB internal protection. It doesn't make any sense to keep it.

I am confused. Now, do you agree that mu lock is preventing the db to be closed when someone is reading it?

liran-funaro:
> I am confused. Now, do you agree that mu lock is preventing the db to be closed when someone is reading it?

No. After careful inspection of the code, I observed that only the `dbList` lock prevents the DB from being closed while someone is reading it.
This is only because the `dbList` is held for the entire duration of the `Get()` method. But this behavior seems unintended because in all other calls `dbList` is held only for reading the map itself.

I can illustrate how without this `dbList` lock "glitch", a delete is possible during a read:
Thread 1:
call Get()
(lock `dbList`)
db <- fetch the `Db` object from the map
(release `dbList`) 
yield

Thread 2:
Commit
delete DB (lock and release mu in the process)
yield

Thread 1:
lock db.mu
continue uninterrupted albeit DB is already deleted and closed
return leveldb.ErrClosed

liran-funaro:
>I agree. So we can go with sync.Map for dbList and per db mutex for the read/delete/create operations?

@cendhu No. See my comment above about why per DB mutex won't suffice.

cendhu:
I see. Okay. We can do as you mentioned. Move the snapshot part to a different pr and keep this pr for lock related issues.

liran-funaro:
>I see. Okay. We can do as you mentioned. Move the snapshot part to a different pr and keep this pr for lock related issues.

I removed the snapshots from this PR. I'll open another PR for that once this PR is merged.

cendhu (approved):

cendhu (approved):
Going forward, please push a PR to focus on one particular issue. For refactoring such as code deduplication, a new PR can be issued. This helps to quickly do the code review.

PR #493, opened 2023-02-05
Avoid instantiating the default marshaller before each use

liran-funaro:
- Fix typo in class name

### DIFF
--- a/pkg/marshal/proto_json_marshal.go
+++ b/pkg/marshal/proto_json_marshal.go
@@ -14,8 +14,8 @@ type DefaultMarshal struct {
 	marshalOption *protojson.MarshalOptions
 }
 
-func DefaultMarshaler() *DefaultMarshal {
-	return &DefaultMarshal{
+func NewDefaultMarshaller() DefaultMarshal {
+	return DefaultMarshal{
 		marshalOption: &protojson.MarshalOptions{
 			Multiline:       false,
 			AllowPartial:    false,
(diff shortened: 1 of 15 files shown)


pkg/marshal/proto_json_marshal.go
```diff
@@ -27,6 +27,8 @@ func DefaultMarshaler() *DefaultMarshal {
 	}
 }
 
+var DefaultMarshaller = NewDefaultMarshaller()
```
cendhu, line 30: We try to avoid global variables as much as possible. Here, we can use an init() function to initialise a local variable. Then, we can make the DefaultMarshaller() to return the initialised variable.

liran-funaro: Alternatively, we can have each struct that uses the `DefaultMarshaller` have its own instance. Is that also acceptable?

cendhu: I initially thought of that but SignPayload and SendHTTPResponse are just functions. Changing signatures of these would requires changes at the SDK too due to the SignPayload function (if I am not wrong). Hence, init() would solve the initial problem.

liran-funaro: I see your point. Can we avoid the init and simply initialize the local variable statically?

cendhu: Okay. I am fine with that.

liran-funaro: fixed

cendhu (approved):

PR #496, opened 2023-02-06
Update goleveldb version to latest

liran-funaro:
There a many bug fixes since the first release of goleveldb.
One of which, is fixing crashes when using released snapshots.

For the full update list, see: https://github.com/syndtr/goleveldb/compare/v1...master


tock-ibm:
This looks fine to me. Fabric also uses a more recent commit of leveldb beyond the v1.0.0. 
@cendhu your thoughts?

PR #497, opened 2023-02-13
Change docker image to Ubuntu

liran-funaro:
Switching to Ubuntu docker image increases the image size from ~30MB to ~140MB.


tock-ibm:
Maybe we can publish both ubuntu and alpine base images?

PR #499, opened 2023-02-16
sync write to block store

cendhu:


internal/blockstore/commit_and_query.go
```diff
@@ -65,6 +65,8 @@ func (s *Store) Commit(block *types.Block) error {
 		return err
 	}
 
+	fileops.SyncDir(s.fileChunksDirPath)
+
```
tock-ibm, line 69: Why do we need to fsync the dir after every write? that is very expensive and has no relation to the bug. I would remove that.

internal/blockstore/commit_and_query_test.go
```diff
@@ -92,33 +92,10 @@ func TestCommitAndQuery(t *testing.T) {
...
-			blockHeaderBytes, err := proto.Marshal(b.GetHeader())
-			require.NoError(t, err)
-			prevBlockHash, err = crypto.ComputeSHA256Hash(blockHeaderBytes)
-			require.NoError(t, err)
-
-			blockHashes = append(blockHashes, prevBlockHash)
-		}
-
-		assertBlocks := func() {
+		assertBlocks := func(startBlockNum, endBlockNum uint64) {
```
tock-ibm, line 95: The changes made to this test do not cover the bug in my environment. That is, I comment out lines 343-345 that seek back to the previous offset, and the test does not fail.

internal/blockstore/commit_and_query.go
```diff
@@ -338,6 +340,9 @@ func (s *Store) Get(blockNumber uint64) (*types.Block, error) {
 		offSet := s.currentOffset
 		defer func() {
 			s.currentOffset = offSet
+			if _, err = f.Seek(offSet, 0); err != nil {
+				s.logger.Panic(err)
+			}
```
tock-ibm, line 345: This fix is good and addresses the problem.

tock-ibm (approved):

tock-ibm:
@cendhu I approved this PR because we need a quick fix for the bug. However, it is not perfect: the `SyncDir` is redundant, and the test does not cover the bug. Lets fix that is a subsequent PR please.

internal/blockstore/commit_and_query_test.go
tock-ibm: In addition, does this test cover the case of reading across multiple chunks?

